PR #20, opened 2017-05-10
User test files

Parastoo-S:
I wrote some test files for user (in test folder). Also made a minor change in the gem file.

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -11,7 +11,7 @@ class User < ActiveRecord::Base
   accepts_nested_attributes_for :repositories
 
   validates :name,
-    length: { maximum: 50, message: 'Your name mus be less than 50 characters.' }
+    length: { maximum: 50, message: 'Your name must be less than 50 characters.' }
 
   validates :username,
     presence: { message: "Your username is required." },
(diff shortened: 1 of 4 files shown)


test/models/user_test.rb
```diff
@@ -1,7 +1,32 @@
...
 
 class UserTest < ActiveSupport::TestCase
-  # test "the truth" do
-  #   assert true
-  # end
+  
+  test "Name length" do 
+  	user = users(:bob)
+  	user.name = "j"*5
+  	assert_equal(-1 , (user.name).length <=> 50, "Your name must be less than 50 characters.")
```
tahnok: Instead of comparing the return code of `<=>` you can say

`assert user.name.length < 50, "Your name must be less than 50 characters"`

test/models/user_test.rb
```diff
@@ -1,7 +1,32 @@
...
+
+  end
+
+  test "Terms and Conditions" do
+  	user = users(:bob)
+  	user.terms_and_conditions = true
+  	assert true == user.terms_and_conditions ,"You must agree to the terms and conditions"
```
tahnok: This is correct, but you can also do `assert user.terms_and_conditions`

test/models/user_test.rb
```diff
@@ -1,7 +1,32 @@
...
+  end
+
+  test "Allowed characters in the password" do
+  	user = users(:bob)
+  	user.password = "abABbc246dabc"
+  	assert_match(/(?=.*[0-9])(?=.*[a-z])(?=.*[A-Z]).{8,}/, user.password , "Your passwords must have one lowercase letter, one uppercase letter, one number and be eight characters long.")
```
tahnok: nice job figuring out ruby's regex syntax

app/models/user.rb
```diff
@@ -11,7 +11,7 @@ class User < ActiveRecord::Base
   accepts_nested_attributes_for :repositories
 
   validates :name,
-    length: { maximum: 50, message: 'Your name mus be less than 50 characters.' }
```
tahnok: nice catch!

test/models/user_test.rb
```diff
@@ -1,7 +1,64 @@
...
+  test "Allowed characters in the password" do
+  	user = users(:bob)
+  	user.password = "abABbc246dabc"
+    assert user.valid? ,"Your passwords must have one lowercase letter, one uppercase letter, one number and be eight characters long."
+  end   
+
+  test "Presence of password" do
+      user = users(:bob)
+      user.password = "abABbc246dabc"
+      # user.password = nil
```
tahnok: leftover?

PR #21, opened 2017-05-11
Karims branch

KarimChukfeh:
fixed a typo


config/sunspot.yml
```diff
@@ -3,17 +3,17 @@ production:
     hostname: 127.0.0.1
     port: 8983
     log_level: DEBUG
-    path: /solr/default
+    #path: /solr/default
```
tahnok: Why did this change? This is working in production so I'm worried about changing it

Gemfile
```diff
@@ -28,12 +28,12 @@ gem 'sdoc', '~> 0.4.0', group: :doc
 gem 'aws-sdk', '~> 2.0'
 gem 'rubyzip', '~> 1.1', require: 'zip'
 gem 'will_paginate'
+gem 'sunspot_solr'
```
tahnok: This gem is already present on line 24

README.md
```diff
@@ -37,25 +37,7 @@ You should now be able to fire up the rails server with:
...
-
-* Services (job queues, cache servers, search engines, etc.)
-
-* Deployment instructions
-
-* ...
+Don't forget to fire up solr with:
+```bash
+sunspot-solr start
+```
```
tahnok: 👍

test/models/user_test.rb
```diff
@@ -1,7 +1,12 @@
 require 'test_helper'
 
 class UserTest < ActiveSupport::TestCase
-  # test "the truth" do
-  #   assert true
-  # end
+
+  context 'user basic info' do
+    setup do
```
tahnok: `setup` is for sharing code between tests and in general shouldn't contain tests. I would create explicit tests for the email/password stuff

KarimChukfeh: yeah this is a mistake on my part something I started but didnt finish. The user_test should be Parastoo's

PR #22, opened 2017-05-11
Clean karim branch

KarimChukfeh:


sunspot-solr.pid
```diff
@@ -0,0 +1 @@
+7740
```
tahnok: I this supposed to be committed?

Gemfile
```diff
@@ -50,7 +50,7 @@ group :test do
   gem 'database_cleaner', '~> 1.0.1'
   gem 'launchy', '~> 2.3.0'
   gem 'selenium-webdriver', '~> 2.39.0'
+  gem 'shoulda'
```
tahnok: Are you using this gem?

sunspot-solr.pid
KarimChukfeh: the file is generated and the pid changes every time you fire up solr.

Gemfile
KarimChukfeh: unsure if i will be using it or not. it can be removed for now

PR #23, opened 2017-05-15
Model tests

Parastoo-S:
Completed tests for a valid repository

### DIFF
--- a/test/models/upvote_test.rb
+++ b/test/models/upvote_test.rb
@@ -2,21 +2,30 @@ require 'test_helper'
 
 class UpvoteTest < ActiveSupport::TestCase
 
-	test "Comment ID uniqueness" do 
-		upvote = Upvote.new(:user_id => 123123 , :comment_id => 456456)
-		assert upvote.invalid? , "You have already upvoted this comment"
+	test "Comment ID uniqueness" do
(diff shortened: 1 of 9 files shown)


test/models/upvote_test.rb
```diff
@@ -1,7 +1,22 @@
 require 'test_helper'
 
 class UpvoteTest < ActiveSupport::TestCase
-  # test "the truth" do
-  #   assert true
-  # end
+
+	test "Comment ID uniqueness" do 
+		upvote = Upvote.new(:user_id => 123123 , :comment_id => 456456)
```
tahnok: Instead of using the ID directly here you can get rails to do this for you!

You have to pull up the user
```
user = users(:bob) #the bob key comes from line 3 of users.yml
```

and then do 
```
Upvote.new(:user => user, comment_id => 456456)
```

You can do the same for comment if you change `upvote.yml` so that the `comment_id` points to a valid `comment`

Parastoo-S: I realized I only needed to check for comment ID's uniqueness and the user_id was only to test for presence. I made changes the "comment_id uniqueness" test I will push. please let me know if I understood what you meant.

PR #26, opened 2017-05-24
Staff dashboard implementation

KarimChukfeh:


app/controllers/staff_dashboard_controller.rb
```diff
@@ -0,0 +1,16 @@
+class StaffDashboardController < ApplicationController
+  before_action :current_user
+  before_action :ensure_admin
```
tahnok: you can do this in one line with `before_action :current_user, :ensure_admin`

spec/controllers/staff_dashboard_controller_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'spec_helper'
```
tahnok: you can delete this file and staff_dashboard_helper_spec.rb

sunspot-solr.pid
```diff
@@ -0,0 +1 @@
+5956
```
tahnok: Probably don't want to commit this, I added a PR so it won't happen again

test/controllers/staff_dashboard_controller_test.rb
```diff
@@ -0,0 +1,23 @@
+require 'test_helper'
+
+class StaffDashboardControllerTest < ActionController::TestCase
+=begin
```
tahnok: How come this is commented out? It seems to run ok on my machine

KarimChukfeh: I get assert_response :found as true for the non-admin fixture (aka bob). As in it redirects to /staff_dashboard instead of root_path

tahnok: That sounds like a bug then, or maybe the response is `:found` but the path is right?

KarimChukfeh: no i print request.original_url and it goes to /staff_dashboard when it shouldn't. check most recent commit

PR #27, opened 2017-05-24
Translations

Parastoo-S:
Translated the sign up page

### DIFF
--- a/app/views/repositories/new.html.erb
+++ b/app/views/repositories/new.html.erb
@@ -58,7 +58,7 @@
     <br>
     <br>
     <br>
-    <%= f.submit "Create Project/CRÉER TON PROJET", class: "save-button", id: "waiting-save-button" %>
+    <%= f.submit "Create Project/Créer ton Projet", class: "save-button", id: "waiting-save-button" %>
     <span id="status-save"></span>
 
   <% end %>
--- a/app/views/repositories/new.html.erb
+++ b/app/views/repositories/new.html.erb
@@ -1,11 +1,11 @@
 <% provide(:title, "New Project") %>
 <section id="new-repo">
-    <div class="title">New Project</div>
+    <div class="title">New Project/Nouveau Projet</div>
 
   <%= form_for @repository, url: {controller: "repositories", action: "create"}, html: { multipart: true } do |f| %>
   <br>
-  <%= f.label :title, "Title" %>
-    <div>Great project titles are short and memorable.</div>
+  <%= f.label :title, "Title/Titre" %>
+    <div>Great project titles are short and memorable./Des titres de projets génial sont court et mémorable.</div>
     <%= f.text_field :title, class: 'repo' %>
     <br>
     <%= f.label :description %>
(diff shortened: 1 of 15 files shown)


app/views/users/new.html.erb
```diff
@@ -3,47 +3,47 @@
...
+    <%= f.label :password_confirmation, "Password Confirmation/ Confirmation de Mot de Passe" %>
     <%= content_tag :span, @new_user.errors[:password_confirmation][0], class: 'form-error' %>
     <%= content_tag :span, @new_user.errors[:password][1], class: 'form-error',
         style: 'display: block;' if @new_user.errors[:password][1] %>
     <%= f.password_field :password_confirmation %>
 
     <div id="terms_and_conditions_box" >
       <%= f.check_box :terms_and_conditions, :style => "float:right;" %>
-      <%= f.label :terms_and_conditions, :class => "checkbox inline" do "I agree to the #{link_to "terms and condition", "terms_and_conditions.pdf", target: "blank"}".html_safe end %>
+      <%= f.label :terms_and_conditions, :class => "checkbox inline" do "I agree to the #{link_to "terms and condition", "terms_and_conditions.pdf", target: "blank"}", "J\'accepte les #{link_to "termes et conditions", "terms_and_conditions.pdf", target: "blank"}".html_safe end %>
```
tahnok: This doesn't quite work the way you want, you need to have it all in the same string, no "," between

app/views/repositories/new.html.erb
```diff
@@ -1,64 +1,64 @@
...
-    <div>Choose how you want your project to be used by others.</div>
+    <%= f.label :license "License/Licence" %> 
+    <div>Choose how you want your project to be used by others./Choisi comment tu veux que ton projet soit utilisé par les autres</div>
     <%= f.select :license, @repository.class.license_options %>
 
     <br>
     <br>
     <br>
-    <%= f.submit "Create Project", class: "save-button", id: "waiting-save-button" %>
+    <%= f.submit "Create Project/CRÉER TON PROJET", class: "save-button", id: "waiting-save-button" %>
```
tahnok: This should probably be lower case

app/views/users/new.html.erb
Parastoo-S: ooh ok I'll fix it

app/views/users/new.html.erb
```diff
@@ -3,47 +3,47 @@
 
   <div class="signup-breadcrumb">
      <div style="background-color:#eee;">
-         Create Account
+         Create Account/Créer le Compte 
      </div>
-     <div>
-         Additional Information
+     <div style= "padding-top: 3% ; padding-bottom: 4%">
```
tahnok: It's generally a good idea to keep styles out of HTML like this because it's harder to re-use. Can you add a class to this div and move the padding stuff into `users.scss`?

Parastoo-S: yes for sure :)

tahnok:
Great! Time to merge

PR #30, opened 2017-05-25
User controller test

KarimChukfeh:


test/controllers/users_controller_test.rb
```diff
@@ -1,39 +1,155 @@
...
+                terms_and_conditions: true,
+                password: "Password1"}
+    assert_response :found, "\nFailed at creating Sam"
+    assert User.exists?(username: "sam"), "\nFailed at saving Sam"
   end
 
-  test "should get new" do
+  ##########
+  #new tests
+  test "new succeeds if user is not signed in and redirects to home if user is" do
```
tahnok: This should be 2 separate tests, one for signed in and one for not signed in

test/controllers/users_controller_test.rb
```diff
@@ -1,39 +1,155 @@
...
+                terms_and_conditions: true,
+                password: "Password1"}
+    assert_response :found, "\nFailed at creating Sam"
+    assert User.exists?(username: "sam"), "\nFailed at saving Sam"
+    @user = User.find_by(username: "sam")
+    assert_nil User.find_by(username: "sam").faculty
+    patch :additional_info, username: "sam", user: {faculty: "engineering"}
+    assert_equal "engineering", User.find_by(username: "sam").faculty
+    assert_redirected_to settings_profile_path
+    assert_response :found
```
tahnok: Don't need this extra `:found` if you were redirected since you check `assert_redirected_to`

test/controllers/users_controller_test.rb
```diff
@@ -1,39 +1,155 @@
...
+    @user = users(:bob)
+    get :additional_info, username: @user.username, user: {}
+    assert_response :found
   end
 
-  test "should get update" do
-    get :update
-    assert_response :success
+  test "should be able to patch additional_info" do
+    post :create, user: {
```
tahnok: It's probably simpler to use an existing user like `:bob` rather than creating a new one.

Creating a new user adds extra steps to the test that might go wrong

test/controllers/users_controller_test.rb
```diff
@@ -1,39 +1,155 @@
...
+    assert_redirected_to settings_profile_path
+    assert_response :found
   end
 
-  test "should get show" do
-    get :show
-    assert_response :success
+  ##########
+  #update tests
+  test "user should be able to update profile with patch" do
```
tahnok: This looks like the test above, perhaps they can be combined?

test/controllers/users_controller_test.rb
```diff
@@ -1,39 +1,155 @@
...
+                password: "Password1"}
+    assert_response :found, "\nFailed at creating Sam"
+    assert User.exists?(username: "sam"), "\nFailed at saving Sam"
+    @user = User.find_by(username: "sam")
+    @oldpass = @user.password
+    post :change_password, username: @user.username,
+      user: {old_password: "Password1", password: "Password2", password_confirmation: "Password2"}
+    @newpass = User.find_by(username: "sam").password
+    assert_equal 'Password changed successfully', flash[:notice]
+    assert @oldpass != @newpass
```
tahnok: you can use `assert_not_equal`

test/controllers/users_controller_test.rb
KarimChukfeh: it's testing different requests though

tahnok:
👍 Merging time!

PR #36, opened 2017-05-31
Clean getting rid of solr

KarimChukfeh:


app/models/repository.rb
```diff
@@ -11,20 +11,20 @@ class Repository < ActiveRecord::Base
...
-            end
-    text    :equipments do
-              equipments.map { |equipment| equipment.name }
-            end
-    integer :like
-    integer :make
-    time    :created_at
-    time    :updated_at
-  end
+  # searchable do
```
tahnok: You can delete this

Gemfile.lock
```diff
@@ -219,9 +218,18 @@ GEM
     responders (2.4.0)
       actionpack (>= 4.2.0, < 5.3)
       railties (>= 4.2.0, < 5.3)
-    rsolr (2.0.1)
-      builder (>= 2.1.2)
-      faraday
+    rspec-core (2.14.8)
```
tahnok: Can you run bundle install again on this branch? I am confused as to why `rspec` is back in this file. IT should be auto generated from `Gemfile`

app/views/repositories/show.html.erb
```diff
@@ -10,12 +10,15 @@
 
   <section id="repository-photos">
 
-    <div id="show-photo">
-      <%= image_tag @photos.first.try(:image).try(:url), data: { width: @photos.first.width, height: @photos.first.height} %>
-    </div>
+
+    <% if false %>
```
tahnok: Is this left over?

app/views/search/category.html.erb
```diff
@@ -18,21 +18,24 @@
   <% @repositories.each_with_index do |repository, i| %>
     <div class="repository-container">
 
-      <div class="repo-display-wrapper" <%= "style=background-image:url('#{@photos[repository.id].image.url}')"  %> >
+      <% if false %>
```
tahnok: And this?

app/views/search/search.html.erb
```diff
@@ -14,27 +14,30 @@
     </nav>
   </section>
   <span class='found'>
-    <%= "#{@repositories.total_entries} repositories were found" %>
+    <%= "#{Repository.where("title LIKE ?", "%#{params[:q]}%").count} repositories were found" %>
```
tahnok: Probably easier to use `@repositories.count`

config/deploy.rb
```diff
@@ -43,51 +43,10 @@
...
-  before 'deploy:starting',  'solr:stop'
-  after  'deploy:finishing', 'solr:start'
-end
-
 # Default value for :linked_files is []
 append :linked_files, 'config/secrets.yml'
 
 
-# Default value for linked_dirs is []
-append :linked_dirs, 'log', 'tmp/pids', 'tmp/cache', 'tmp/sockets', 'public/system', 'solr/data', 'solr/pids'
```
tahnok: you need to leave the rest of these in, just remove `solr/data` and `solr/pids`

KarimChukfeh:
regarding the <% if false %> I'll be working on that as soon as I'm done with training and certification. I opened an issue for it.

Gemfile.lock
tahnok: Ya, you need to run `bundle install` and then commit the changes to `Gemfile.lock` so that these disappear

app/controllers/search_controller.rb
```diff
@@ -9,37 +9,38 @@ def explore
 
   def search
   	sort_arr = sort_order
-  	@repositories = Repository.search do
-	    keywords params[:q] 
+  	@repositories = Repository.where("title LIKE ?
+                                  OR description LIKE ?
+                                  OR user_username LIKE ?
+                                  OR category LIKE ?",
+                                  "%#{params[:q]}%", "%#{params[:q]}%", "%#{params[:q]}%", "%#{params[:q]}%").paginate(:per_page=>12,:page=>params[:page]) do
```
tahnok: Do you need the `paginate` params on the end of this? It looks like those are added the query inside the block that follows (line 18). It feels redundant to have both

PR #42, opened 2017-06-08
Clean training and cert

KarimChukfeh:


app/controllers/staff_dashboard_controller.rb
```diff
@@ -1,15 +1,3 @@
-class StaffDashboardController < ApplicationController
-  before_action :current_user, :ensure_admin
-
-  def index
```
tahnok: Where did this go?

app/helpers/application_helper.rb
```diff
@@ -14,7 +14,11 @@ def github_client
   def admin?
     @user.role.eql?("admin")
   end
-  
+
+  def staff?
```
tahnok: I think this method should probably live in the `user` model so you can do `user.staff?` or `user.admin?`

app/views/admin/users/show.html.erb
```diff
@@ -74,7 +74,7 @@
       <div>This user is currently certified to use the following pieces of equipment:</div>
       <div id="cert-container">
         <% @certifications.each do |certification| %>
-          <span class="certification"><%= certification.name %></span>
+          <span class="certification"><%= certification.training %></span>
```
tahnok: Do you need to say `certification.training.name`?

app/views/layouts/staff_dashboard.html.erb
```diff
@@ -0,0 +1,5 @@
+<!DOCTYPE html>
+<html>
+<head>
```
tahnok: Is this used?

app/views/staff/index.html.erb
```diff
@@ -0,0 +1,141 @@
+<!DOCTYPE html>
```
tahnok: You shouldn't need these extra html/head tags here. The layout file that you are using should include them

app/views/staff/index.html.erb
```diff
@@ -0,0 +1,141 @@
...
+
+      <div class="locations">
+      <table class="locations">
+        <% if params[:location].nil? %>
+            <td class="current">
+        <% else %>
+            <td>
+        <% end %>
+          <%= link_to "All locations", :sort => params[:sort], :direction => params[:direction], :p => params[:p] %></td>
+          <% PiReader.uniq.pluck(:pi_location).sort_by!{ |m| m }.each do |pi_location| %>
```
tahnok: Why `uniq`? Also you don't need to do `sort_by!`, you should let the database do the sorting and use `order`

app/views/staff/index.html.erb
```diff
@@ -0,0 +1,141 @@
...
+    <div class="search-form">
+    <%= form_tag search_admin_users_path, method: :get do %>
+      <%= select_tag 'filter', options_for_select(["Name", "Email", "Username"]), prompt:'Search by...' %>
+      <%= text_field_tag 'q', nil, placeholder: 'Search users...', class:'profile-text' %>
+      <%= submit_tag "Search", class: 'save-button' %>
+    <% end %>
+  </div>
+  <br>
+  <div>
+    <% if params[:p] == "signed_in_users" %>
```
tahnok: What is this `:p` variable? Can it have a clearer name?

db/migrate/20170608155514_add_training_session_id_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddTrainingSessionIdToUsers < ActiveRecord::Migration
+  def change
+    add_column :users, :training_session_id, :string
```
tahnok: I'm not sure this is going to work the way you want. This means that each user can only ever be in a single training session. I'm also not sure that having training information exist on the user is good design because it breaks encapsulation. The training session model depends on users, but the user doesn't need to have training sessions.

If you want to track the users in a training session you will probably need to create a [Join table](https://en.wikipedia.org/wiki/Associative_entity) that only has 2 columns: user_id and training_session_id OR you can use [has_many :through](http://guides.rubyonrails.org/association_basics.html#the-has-many-through-association) and use certification as the 'join table'

test/controllers/admin/settings_controller_test.rb
```diff
@@ -0,0 +1,36 @@
+require 'test_helper'
+
+class Admin::SettingsControllerTest < ActionController::TestCase
+
+  setup do
+    session[:user_id] = User.find_by(username: "adam").id
+    session[:expires_at] = "Sat, 03 Jun 2020 05:01:41 UTC +00:00"
+    @staff = User.find_by(username: "adam")
+    @request.env['HTTP_REFERER'] = admin_settings_url
```
tahnok: How come you need this?

test/controllers/admin/settings_controller_test.rb
```diff
@@ -0,0 +1,36 @@
+require 'test_helper'
+
+class Admin::SettingsControllerTest < ActionController::TestCase
+
+  setup do
+    session[:user_id] = User.find_by(username: "adam").id
```
tahnok: Since `adam` is a fixture you can access it like `users(:adam)`

test/controllers/admin/settings_controller_test.rb
```diff
@@ -0,0 +1,36 @@
...
+  test "admin can add training" do
+    post :add_training, training_name: "soldering_1"
+    assert_equal flash[:notice], "Training added successfully!"
+    assert Training.find_by(name: "soldering_1").present?
+  end
+
+  test "admin can rename training" do
+    patch :rename_training, training_name: "lathe_1", training_new_name: "lathe_2"
+    assert_equal flash[:notice], "Training renamed successfully"
+    assert !Training.find_by(name: "lathe_1").present?
```
tahnok: Instead of `assert !something` you can use `refute something` to make it more obvious you are checking the negative condition

app/controllers/staff_dashboard_controller.rb
KarimChukfeh: will put it back

test/controllers/admin/settings_controller_test.rb
KarimChukfeh: True I just feel more confident when the test pass by querying the actual database

test/controllers/admin/settings_controller_test.rb
KarimChukfeh: Without `@request.env['HTTP_REEFER']` i get: 
> ActionController::RedirectBackError: No HTTP_REFERER was set in the request to this action, so redirect_to :back could not be called successfully. If this is a test, make sure to specify request.env["HTTP_REFERER"].

db/migrate/20170608155514_add_training_session_id_to_users.rb
KarimChukfeh: Oh yeah this makes sense. `training_session_id` seems unnecessary since it's only the last training session the user was in.

app/helpers/application_helper.rb
KarimChukfeh: done

app/controllers/staff_dashboard_controller.rb
KarimChukfeh: done

app/views/admin/users/show.html.erb
KarimChukfeh: fixed

app/views/layouts/staff_dashboard.html.erb
KarimChukfeh: removed

app/views/staff/index.html.erb
KarimChukfeh: to be honest I'm not entirely sure how pagination works

test/controllers/admin/settings_controller_test.rb
KarimChukfeh: done

db/migrate/20170608155514_add_training_session_id_to_users.rb
KarimChukfeh: done

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -0,0 +1,130 @@
+class Staff::TrainingSessionsController < StaffAreaController
+  before_action :load_user, only: [:show, :edit, :update]
+
+  layout 'staff_area'
+
+  def index
```
tahnok: I find this method really really confusing, maybe we can thin it out. Do we need all the pagination and sorting stuff? Maybe that can be added in another PR after this one is merged?

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -0,0 +1,130 @@
...
+  def create
+    staff = current_user
+    training_session = TrainingSession.new(training_session_params)
+    if training_session.save
+      flash[:notice] = "Training session created succesfully"
+    end
+    redirect_to staff_training_sessions_url
+  end
+
+  def rename_training_session
```
tahnok: This doesn't look like renaming, this looks like setting which training the current training session is for

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -0,0 +1,130 @@
...
+      if training_session.save
+        flash[:notice] = "Training session renamed succesfully"
+      end
+    end
+    redirect_to :back
+  end
+
+  def reschedule_training_session
+    staff = current_user
+    training_session = TrainingSession.find_by(training_session_params)
```
tahnok: Most of these `find_by` should be replaced with a `TrainingSession.find(id)` and assigning the current training session should in a before action like in [admin/users_controller](https://github.com/uOttawa-Makerspace/MakerSpaceRepo/blob/master/app/controllers/admin/users_controller.rb#L129)

app/controllers/staff/training_sessions_controller.rb
KarimChukfeh: to be honest this is reused code. Yeah I'll work on replacing it in another PR.

app/controllers/staff/training_sessions_controller.rb
KarimChukfeh: yeah i should probably give this a different name

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -0,0 +1,130 @@
...
+    if params['training_session_new_time'].present?
+      training_session.timeslot = params['training_session_new_time']
+      if training_session.save
+        flash[:notice] = "Training session rescheduled succesfully"
+      end
+    end
+    redirect_to :back
+  end
+
+  def delete_training_session
```
KarimChukfeh: hmm
>ActionController::UrlGenerationError: No route matches {:action=>"destroy", :controller=>"staff/training_sessions", :training_session=>{:training_id=>"1", :timeslot=>"2018-06-02T02:01:41+00:00", :user_id=>"777"}}
    test/controllers/staff/training_sessions_controller_test.rb:68:in `block in <class:TrainingSessionsControllerTest>'

app/controllers/staff/training_sessions_controller.rb
KarimChukfeh: done

app/controllers/staff/training_sessions_controller.rb
KarimChukfeh: i don't get this :confused:

app/controllers/staff/training_sessions_controller.rb
tahnok: Do you mean replacing it later and leaving it as is or cleaning up what's here and adding it later?

KarimChukfeh: I mean leaving it as is and replacing it later

app/controllers/staff/training_sessions_controller.rb
tahnok: I wrote up a longer message in slack, but instead of sending the 3 parameters to figure out which training session to update, you should use the ID of the training_session once it has been created. Once you pass in the ID you can just do `TraingSession.find(params[:id])` which is how the the admin/users_controller works

app/controllers/staff/training_sessions_controller.rb
tahnok: I'd actually prefer if you removed it. Confusing code we don't need right now is worse than no code

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -1,116 +1,63 @@
...
+      flash[:alert] = "Something went wrong. Please try again."
+      redirect_to :back
     end
-    redirect_to staff_training_sessions_url
   end
 
-  def rename_training_session
-    staff = current_user
-    training_session = TrainingSession.find_by(training_session_params)
+  def update
```
tahnok: You may be able to replace much of this with just `@current_training_session.update(training_session_params)` or something like that

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -1,116 +1,63 @@
...
-    training_session = TrainingSession.find_by(training_session_params)
     if params['training_session_graduates'].present?
       params['training_session_graduates'].each do |graduate|
-         if training_session.users.include? User.find(graduate)
-           unless User.find(graduate).certifications.include? Certification.find_by(training: Training.find(training_session_params['training_id']).name)
-             certification = Certification.new(user_id: graduate, trainer_id: staff.id, training: Training.find(training_session_params['training_id']).name)
+         if @current_training_session.users.include? User.find(graduate)
+           unless User.find(graduate).certifications.include? Certification.find_by(training: @current_training_session.training.name)
+             certification = Certification.new(user_id: graduate, trainer_id: staff.id, training: @current_training_session.training.name)
              certification.save
```
tahnok: probably need to check the return value of this `save` since it might fail

config/routes.rb
```diff
@@ -115,11 +115,8 @@
 
     resources :training_sessions do
       collection do
```
tahnok: This `collection` method is supposed to be used to add methods to a collection of resources (like training_sessions). Usually that would be something like search.

Instead, you should use `member` (see [rails routing guide](http://guides.rubyonrails.org/routing.html#adding-more-restful-actions)). This will mean that the url will include the id in like `/staff/training_sessions/IDHERE/certifiy_trainees` instead of the way it is now where it is `/staff/training_sessions/certifiy_trainees`

app/controllers/staff/training_sessions_controller.rb
KarimChukfeh: i think i did this already :o

KarimChukfeh: oh nvm i get it

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+      redirect_to :back
+      flash[:alert] = "Something went wrong. Please try again."
+    end
+  end
+
+  def update
+    #binding.pry
+    @current_training_session.update(changed_params)
+    if params['users'].present?
+      JSON.parse(params['users']).each do |new_user|
```
tahnok: If you pass the parameters in the right format you shouldn't need this JSON parsing. Something like [this stack overflow](https://stackoverflow.com/a/23066966)

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+             end
+           end
+         end
+       end
+       flash[:notice] = "Users certified successfuly"
+     end
+     redirect_to :back
+  end
+
+  def delete_training_session
```
tahnok: This should be destroy, not sure why it wasn't working before but it's worth spending more time trying to get right

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+    end
+    redirect_to :back
+  end
+
+  def certify_trainees
+    staff = current_user
+    if params['training_session_graduates'].present?
+      params['training_session_graduates'].each do |graduate|
+         if @current_training_session.users.include? User.find(graduate)
+           unless User.find(graduate).certifications.include? Certification.find_by(training: @current_training_session.training.name)
```
tahnok: This kind of uniqueness checking should live in the model. Something like this: https://stackoverflow.com/questions/8965089/validate-uniqueness-of-in-association

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -0,0 +1,91 @@
...
+             certification = Certification.new(user_id: graduate, trainer_id: staff.id, training: @current_training_session.training.name)
+             unless certification.save
+               flash[:alert] = "%{grad}'s certification not saved properly!" % { :grad => graduate.username }
```
tahnok: I think this will only show the last error which is not what you wanted

app/controllers/staff_controller.rb
```diff
@@ -0,0 +1,7 @@
+class StaffController < StaffAreaController
+
+  def index
+    @users = User.all
```
tahnok: This should be limited to the last 10 users otherwise this will be a HUGE problem in production where there are 100s of users

app/models/certification.rb
```diff
@@ -1,3 +1,10 @@
 class Certification < ActiveRecord::Base
   belongs_to :user
+  belongs_to :training_session
+
+  validates :user, presence: true
+  validates :trainer_id, presence: true
+  validates :training, presence: true
+  scope :show_options, -> { order("lower(training) ASC").all }
```
tahnok: Where is this used?

app/views/staff/index.html.erb
```diff
@@ -0,0 +1,81 @@
...
+            singout_time.strftime("%B %d, %Y")
+          end
+        %></td>
+
+        <td class="location-cell"><%=
+          user.location
+        %></td>
+
+        <td class="certifications-cell"><%
+          if user.certifications.first.present?
```
tahnok: Why do you need this check? If there aren't any certifications you won't loop over anything

app/models/certification.rb
```diff
@@ -1,3 +1,10 @@
 class Certification < ActiveRecord::Base
   belongs_to :user
+  belongs_to :training_session
+
+  validates :user, presence: true
+  validates :trainer_id, presence: true
```
tahnok: This might not be necessary. The certification comes from a training_session which has a trainer already.

app/views/staff/index.html.erb
```diff
@@ -0,0 +1,81 @@
...
+            user.certifications.each do |certification|
+            %>
+            <div
+             class="cert-name"
+             style="display:inline" >
+             <div class="cert-holder" style="display:inline">
+              <%= certification.training %> |
+              <div id="cert-details" style="display:inline">
+                <%= certification.training %> <br>
+                <%= User.find(certification.trainer_id).name %> <br>
```
tahnok: I would add a method to training to do this for you. You could do `certification.training_session.trainer.name` which is much clearer

app/views/staff/index.html.erb
```diff
@@ -0,0 +1,81 @@
...
+            end
+          end
+          %>
+      </td>
+
+      </tr>
+    <% end %>
+    </table>
+    <br>
+    <input type="hidden", name="training_session[user_id]", value="<%=current_user.id%>"/>
```
tahnok: I wouldn't include this as an input in the form, I would fetch the current user in the controller instead

app/views/staff/index.html.erb
```diff
@@ -0,0 +1,81 @@
...
+    <input type="hidden", name="training_session[training_id]", value="<%=Training.first.id%>"/>
```
tahnok: Is this just for testing?

app/views/staff/index.html.erb
```diff
@@ -0,0 +1,81 @@
...
+    <input type="hidden", name="training_session_users", value=""/>
+    <script>
+      function check_users(){
```
tahnok: You may be able to replace all this javascript with something like the code in this stack overflow: https://stackoverflow.com/questions/21568957/using-checkboxes-to-submit-values-to-an-array

Basically use the rails form helpers to generate a checkbox

app/views/staff/training_sessions/show.html.erb
```diff
@@ -0,0 +1,178 @@
...
+      Other Users
+      <tr>
+        <th>Add</th>
+        <th>Username</th>
+        <th>Last Signed In</th>
+        <th>Location</th>
+        <th>Certification</th>
+      </tr>
+
+    <% User.all.each do |user| %>
```
tahnok: This is OK for now, but we will need something better in a new PR.

test/controllers/admin/settings_controller_test.rb
```diff
@@ -0,0 +1,11 @@
+require 'test_helper'
+
+class Admin::SettingsControllerTest < ActionController::TestCase
```
tahnok: Are there tests missing from here? Maybe you can add some for the settings controller in a new PR

test/controllers/admin/users_controller_test.rb
```diff
@@ -0,0 +1,18 @@
...
+class Admin::UsersControllerTest < ActionController::TestCase
+  setup do
+    session[:user_id] = User.find_by(username: "adam").id
+    session[:expires_at] = "Sat, 03 Jun 2020 05:01:41 UTC +00:00"
+    #@staff = User.find_by(username: "adam")
+    @request.env['HTTP_REFERER'] = admin_users_url
+  end
+
+
+  test "admin can change user role (admin > staff > regular_user)" do
```
tahnok: 🆗 👌

test/controllers/staff_dashboard_controller_test.rb
```diff
@@ -2,16 +2,16 @@
...
-     get :index
-     assert_response :success
-   end
-
-   test "regular users are redirected to home" do
-     session[:user_id] = users(:bob).id
-     get :index
-     assert_redirected_to root_path
-   end
+  #  test "staff succeed at loading Staff Dashboard" do
```
tahnok: Seems like this shouldn't be commented out?

app/controllers/admin/trainings_controller.rb
```diff
@@ -0,0 +1,44 @@
+class Admin::TrainingsController < AdminAreaController
+  layout 'admin_area'
+
+  def add_training
```
KarimChukfeh: done

app/controllers/admin/trainings_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    if params[:training_name].present? && !Training.find_by(name: params[:training_name]).present?
```
KarimChukfeh: done

app/controllers/staff/training_sessions_controller.rb
KarimChukfeh: done

app/models/certification.rb
KarimChukfeh: i think this slipped while trying things

app/views/staff/index.html.erb
KarimChukfeh: done

test/controllers/admin/settings_controller_test.rb
KarimChukfeh: this contained trainings tests before they were moved to their own controller

app/views/staff/index.html.erb
KarimChukfeh: done

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -1,5 +1,7 @@
 class Staff::TrainingSessionsController < StaffAreaController
-  before_action :current_training_session, only: [:show, :edit, :update, :certify_trainees, :delete_training_session]
+
+  skip_before_action :training_session_params, only: [:create]
```
tahnok: I don't see where this before action is added so that you need to skip it

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -13,22 +15,21 @@ def new
   end
 
   def create
-    @training_session = TrainingSession.new(training_session_params)
+    @new_training_session = TrainingSession.new(user_id: current_user.id, training_id: Training.all.first.id)
```
tahnok: you should use a param here

KarimChukfeh: done

app/controllers/staff/training_sessions_controller.rb
KarimChukfeh: done

app/views/staff/index.html.erb
KarimChukfeh: yeah it's removed now

app/views/staff/index.html.erb
KarimChukfeh: done

app/views/staff/index.html.erb
KarimChukfeh: done

app/views/staff/training_sessions/show.html.erb
```diff
@@ -0,0 +1,130 @@
+<section id ="page-contents" >
+
+<div>Trainer: <%= User.find(@current_training_session.user_id).name%></div><br>
```
tahnok: This user should be assigned in an instance variable in the controller

test/controllers/staff_dashboard_controller_test.rb
tahnok: Is this still commented out for a reason?

app/controllers/staff_dashboard_controller.rb
```diff
@@ -1,15 +1,7 @@
...
   def index
-  end
-
-  private
-
-  def ensure_admin
-    unless admin?
-      redirect_to '/' and return
-    end
+    redirect_to staff_dashboard_index_path
```
tahnok: Is this an infinite redirect loop? Isn't this index action `staff_dashboard_index_path`

app/models/training_session.rb
```diff
@@ -0,0 +1,15 @@
+class TrainingSession < ActiveRecord::Base
+  belongs_to :training
+  belongs_to :user
+  has_and_belongs_to_many :users, :uniq => true
+  has_many :certifications
+
+  validates :training, presence: { message: "A training subject is required"}
+  validates :user, presence: { message: "A trainer subject is required"}
```
Parastoo-S: I think "subject" should be removed.
"A trainer is required"

KarimChukfeh: good catch!

db/schema.rb
```diff
@@ -11,18 +11,20 @@
...
   enable_extension "plpgsql"
 
   create_table "categories", force: :cascade do |t|
     t.integer  "repository_id"
     t.string   "name"
-    t.datetime "created_at",    null: false
-    t.datetime "updated_at",    null: false
+    t.datetime "created_at",         null: false
+    t.datetime "updated_at",         null: false
+    t.integer  "category_option_id"
```
tahnok: I think this snuck in

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -0,0 +1,96 @@
...
+    if @current_training_session.save
+      flash[:notice] = "Training session updated succesfully"
+    else
+      flash[:alert] = "Something went wrong, please try again"
+    end
+    redirect_to :back
+  end
+
+  def certify_trainees
+    staff = current_user
```
tahnok: This looks unused?

PR #44, opened 2017-06-08
Reports

Parastoo-S:
***THIS PR SHOULD BE DELETED***


app/controllers/report_generator_controller.rb
```diff
@@ -0,0 +1,27 @@
+class ReportGeneratorController < ApplicationController
```
tahnok: You will want to be sure to inherit from `AdminAreaController` so that only admins can access this page

Parastoo-S: oh okay. I will

Parastoo-S:
I think I messed up my branch again :'( I used git rebase yesterday to update my branch with the master but now all the commits and file changes have been added:(

tahnok:
You can close this PR if you aren't going to use it

PR #48, opened 2017-06-14
new wearables image

KarimChukfeh:


tahnok:
No need to keep the old images, they are in git forever and we can pull them out again if needed

PR #49, opened 2017-06-19
Rake task for ta c

KarimChukfeh:


tahnok:
What's the difference between this and https://github.com/uOttawa-Makerspace/MakerSpaceRepo/pull/50 ?

PR #50, opened 2017-06-19
Rake task for sending emails

KarimChukfeh:


app/mailers/msr_mailer.rb
```diff
@@ -26,4 +26,10 @@ def send_report(email, csv1, csv2, csv3, csv4, csv5)
 
 	    mail(to: email, subject: 'Monthly Report')
 	end
+
+	def tac_reminder_email(email)
+    mail(to: email,
```
tahnok: Formatting here is realy weird

app/views/msr_mailer/tac_reminder_email.html.erb
```diff
@@ -0,0 +1,3 @@
+<%= image_tag("EmailHeader.jpg", width: '100%', position: 'relative') %>
+<h1>If you don't sign the terms and conditions you can't access your account and your account will be suspended!</h1>
```
tahnok: We probably need to check with Zaineb on how to word this

tahnok:
This is almost ready to go, you just need to update the wording

PR #56, opened 2017-07-04
Fixing search 2

KarimChukfeh:


tahnok:
I spent some time looking at the actual data in production and the current state of affairs is very not good.

Previously, a repository had a column in the db called 'category' and that's how projects got categorized. The only values that are there now are: `["Bio-Medical", "Wearables", "Internet of Things", "Mobile", "3D-Model", "Virtual Reality", nil]`

After some changes last summer, each project now has_many categories, which is a new model/table in the database. Each category has a name and the only names currently are: ` ["Health Sciences", "Wearable", "Internet of Things", "Mobile Development", "Virtual Reality", "Other Projects", "Course-related Projects", "uOttawa Team Projects", "undefined"]`. These names come from another model called CategoryOption which limits which names can go in category.

Finally, the `static_pages/home.html.erb` links to the search controller with these slugs: `wearable, mobile-development, uottawa-team-projects, health-sciences, internet-of-things, other-projects, virtual-reality, course-related-projects`

As you can see, there does not seems to be a straightforward mapping of the slugs in the home page to the 2 ways of categorizing things that currently exist. This is the core of our problem.

I'm not sure what the right solution is. I suspect that you will have to add 2 hashes that map the slug from the `home.html.erb` to either the old category field on the repository or to the new category model.

Something like:

```
SLUG_TO_OLD_CATEGORY = {
'internet-of-things' => 'Internet of Things',
}

SLUG_TO_CATEGORY_MODEL = {
 'internet-of-things' => 'Internet of Things',
'course-related-projects' => 'Course-related Projects' #this is the really weird one
}

def category
 @repositories
 if category = SLUG_TO_OLD_CATEGORY[param[:slug]]
 @repositories += Repository.where(category: category)
end

if name = SLUG_TO_CATEGORY_MODEL[param[:slug]]
 @repositories += Category.where(name: "foo").includes(:repository).map(&:repository)
end
```

The above approach limits the number of times we hit the database. `Category.all` would load each and every category just to check, where the above will only load the records we care about. The `.includes(:repository)` is a bit of a trick to avoid what's called the [N+1 query problem](http://www.benlcollins.com/sql/fixing-the-n1-problem-in-my-rails-application/)

I'd also like to see you add some tests for this. There's a fairly complex flow here but it should be pretty easy to test

app/controllers/search_controller.rb
```diff
@@ -18,11 +18,12 @@ def search
                                   "%#{params[:q].downcase}%",
                                   "%#{params[:q].downcase}%",
                                   "%#{params[:q].downcase}%")
-    repositories_by_categories = []
-    Category.where('lower(name) LIKE ?', params[:q].downcase).each do |cat|
-      repositories_by_categories << cat.repository
-	  end
-    @repositories = repositories_by_attributes + repositories_by_categories
+    if category = SLUG_TO_OLD_CATEGORY[params[:slug]]
```
tahnok: I don't think this is correct for search since there isn't a param called `:slug`. I'm not sure how to fix the weird category stuff for search honestly

app/models/repository.rb
```diff
@@ -20,6 +20,10 @@ def self.license_options
      "Attribution - Non-Commercial - No Derivatives"]
   end
 
+  def user_username
```
tahnok: WHy did this get added?

test/controllers/search_controller_test.rb
```diff
@@ -1,14 +1,31 @@
...
+  end
+
+  test "Shows repos with new categorizing (aka Repository has_many Category)" do
+    get :category, slug: 'course-related-projects'
+    assert_response :success
+    assert response.body.include? "Repository2"   #Course-related Projects -> course-related-projects
+    refute response.body.include? "Repository1"
+  end
+
+  test "Shows repos with both old and new categorizing" do
```
tahnok: 👍 💯

app/models/repository.rb
KarimChukfeh: because for some reason when I give the first repository fixture `user_username: bob` the following test fails:
`test "Uniqueness of repository title" do 
  repository = repositories(:two)  
  repository.title = "Repository1"
  assert repository.invalid? , "Project title is already in use."` 

even though there's a `user_username` column in Repository. Adding this seem to pass the test as well as render views properly

app/controllers/search_controller.rb
```diff
@@ -10,19 +10,15 @@ def explore
...
                                   "%#{params[:q].downcase}%",
                                   "%#{params[:q].downcase}%",
                                   "%#{params[:q].downcase}%")
-    repositories_by_categories = []
-    Category.where('lower(name) LIKE ?', params[:q].downcase).each do |cat|
-      repositories_by_categories << cat.repository
-	  end
-    @repositories = repositories_by_attributes + repositories_by_categories
+
     @repositories = @repositories.uniq
```
tahnok: You can probably collapse all of these manipulations into 1 assignment which will prevent active record for loading _every_ repository and instead just load the 12 needed for the pagination. Something like

```
@repositories = Repository.where(stuff)distinct.paginate(bla)
```

app/models/repository.rb
tahnok: ah, so that's a valid-ish failure. The test was assuming (incorrectly) that it could just set the title of `fixture(:one)` to the title of `fixture(:two)` and cause a failure. That _was_ the case because the uniqueness check is for uniquness per user_username. After you add user_username, the uniqnuess validation passes. The correct way to fix this isn't to rewrite this method, although I agree it's weird that this exists at all and should be removed later. I've made an issue to track it https://github.com/uOttawa-Makerspace/MakerSpaceRepo/issues/58

app/controllers/search_controller.rb
KarimChukfeh: so i updated this and made it shorted but i'm not sure if the 'stuff' inside where() can become shorter too

wesleyjellis: that's fine, I don't think it can be made any shorter

app/controllers/search_controller.rb
```diff
@@ -4,40 +4,32 @@ class SearchController < SessionsController
...
-    repositories_by_categories = []
-    Category.where('lower(name) LIKE ?', params[:q].downcase).each do |cat|
-      repositories_by_categories << cat.repository
-	  end
-    @repositories = repositories_by_attributes + repositories_by_categories
-    @repositories = @repositories.uniq
-    @repositories.paginate(:per_page=>12,:page=>params[:page]) do
-      order_by sort_arr.first, sort_arr.last
-    end
+                                  "%#{params[:q].downcase}%").uniq
```
wesleyjellis: This should be `distinct` which will make the db do the uniqueness stuff for you

KarimChukfeh: done

app/models/repository.rb
tahnok: Have you looked into removing this method and fixing the tests? #58 is to remove the column

PR #57, opened 2017-07-06
Fixing categories

KarimChukfeh:
I wrote couple migrations to make every Category belongs_to CategoryOption. Further I added a CategoryOption.repositories method. Next I'll add a migration to get of category column in Repository and make these old records use the new Category system.


app/controllers/repositories_controller.rb
```diff
@@ -65,7 +65,7 @@ def create
       create_categories
       create_equipments
       render json: { redirect_uri: "#{repository_path(@user.username, @repository.slug)}" }
-      Repository.reindex
+      # Repository.reindex
```
tahnok: probably worth deleting instead of just commenting. I think this was a solr thing

app/models/category_option.rb
```diff
@@ -1,5 +1,10 @@
 class CategoryOption < ActiveRecord::Base
   belongs_to :admin
-  validates :name, presence: true
+  validates :name, presence: true, uniqueness: true
   scope :show_options, -> { order("lower(name) ASC").all }
+
+  def repositories
```
tahnok: I think you can replace this with `has_many :categories`, `has_many :repositories, through: categories`

db/migrate/20170706152142_add_category_option_id_to_categories.rb
```diff
@@ -0,0 +1,8 @@
+class AddCategoryOptionIdToCategories < ActiveRecord::Migration
+  def change
+    Category.all.each do |cat|
+      cat.category_option_id = CategoryOption.find_by(name: cat.name).id
+      cat.save
```
tahnok: It would be safer to do `save!` which will raise an exception and stop if something happens

db/schema.rb
```diff
@@ -33,9 +34,9 @@
 
   create_table "certifications", force: :cascade do |t|
     t.integer  "user_id"
-    t.string   "name"
-    t.datetime "created_at", null: false
-    t.datetime "updated_at", null: false
+    t.datetime "created_at",          null: false
+    t.datetime "updated_at",          null: false
+    t.integer  "training_session_id"
```
tahnok: You may have to reset your database and then run `rake db:schema:dump` because this shouldn't be here

test/controllers/repositories_controller_test.rb
```diff
@@ -0,0 +1,16 @@
+require 'test_helper'
+
+class RepositoriesControllerTest < ActionController::TestCase
+
+  test "can create new repository" do
```
tahnok: :+1:

app/controllers/repositories_controller.rb
```diff
@@ -174,9 +172,12 @@ def update_files
     end
 
     def create_categories
-      params['categories'].first(5).each do |c|
-        Category.create(name: c, repository_id: @repository.id)
-      end if params['categories'].present?
+      if params['categories'].present?
+        cats = params['categories']
+        cats.each do |cat|
+          Category.create(category_option_id: CategoryOption.find_by(name: String(cat)).id, repository_id: @repository.id)
```
tahnok: Why are you casting cat to a string? The ruby way to do this is `cat.to_s`

db/migrate/20170706152142_add_category_option_id_to_categories.rb
```diff
@@ -0,0 +1,8 @@
+class AddCategoryOptionIdToCategories < ActiveRecord::Migration
+  def change
+    Category.all.each do |cat|
+      cat.category_option_id = CategoryOption.find_by(name: cat.name).id
```
tahnok: I think this might not work properly with the addition of the `name` method to category. Without the manual addition of `name`, `cat.name` will fetch the field in the database. After adding the new `name` method `cat.name` will fetch the category_option and get it's name which will be null here.

I think the solution is to add `name` to category in another PR after this one has been merged and deployed

db/migrate/20170706152142_add_category_option_id_to_categories.rb
```diff
@@ -0,0 +1,8 @@
+class AddCategoryOptionIdToCategories < ActiveRecord::Migration
+  def change
```
tahnok: You should change this from `def change` to `def up` and `def down` like in the examples section of http://api.rubyonrails.org/classes/ActiveRecord/Migration.html

db/migrate/20170706172353_remove_category_column_from_repositories.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveCategoryColumnFromRepositories < ActiveRecord::Migration
+  def change
+    remove_column :repositories, :category, :string
```
tahnok: This needs to be in a separate PR since you have code in here that depends on this

app/controllers/repositories_controller.rb
KarimChukfeh: it didn't work before casting to string but now it does 👍

db/migrate/20170706152142_add_category_option_id_to_categories.rb
KarimChukfeh: done

db/migrate/20170706152142_add_category_option_id_to_categories.rb
KarimChukfeh: Sounds good

db/migrate/20170706151010_add_category_option_to_categories.rb
```diff
@@ -1,5 +1,5 @@
 class AddCategoryOptionToCategories < ActiveRecord::Migration
-  def change
+  def up
```
tahnok: I think you need to have a `down` method too, but it can be empty

db/migrate/20170706172353_remove_category_column_from_repositories.rb
```diff
@@ -0,0 +1,5 @@
+class RemoveCategoryColumnFromRepositories < ActiveRecord::Migration
+  def down
```
tahnok: This should be part of 'up' since it will happen when the migration is 'brought up'. Right now it will only run if you `rake db:rollback` once this has been migrated

tahnok:
Is there something I can do to help you break this down into smaller pieces? Once that happens we can start merging each piece and deploying one by one

tahnok:
Oh, I'm an idiot and didn't see some of your other PRs. NEVERMIND!

PR #66, opened 2017-07-18
validate through the Certification model

KarimChukfeh:
This branch is to implement a solution to duplicate certifications 

The problem is that 

`certifcation` `belongs_to training_session` 

and 

`training_session` `belongs_to training` 

and checking for uniqueness of `training_session` is pointless because a user can be certified to the same `training` but in a different `training_session`.

The current approach is to check every certification that belongs to the user in the certification model before certifying. 

The method is commented out because it depends on the clean_training_and_certification branch.


tahnok:
What's the state of this? I think the dependent branch has merged

PR #67, opened 2017-07-19
Projects

Parastoo-S:
Adding password to private projects


app/views/repositories/password_entry.html.erb
```diff
@@ -0,0 +1,15 @@
...
+
+  <%= password_field_tag :password, nil, placeholder: 'Password' %>
+
+  <%= submit_tag "Sign In/Connexion", class: 'login-button', 'tab-index': "3" %>
+
+<% end %>
+
+
+
+<br>
```
Parastoo-S: This is a temporary view. Will try to figure out how to do it as a pop up.

PR #68, opened 2017-07-20
fix makes

KarimChukfeh:
Makes are broken in development and in production. What seems to have been the issue is two things:

1. the wrong `params` were being handled inside `def create` in the makes controller.
2. `def copy_tags` was creating Tags which is a model that doesn't exist.

The first problem was solved but I'm not sure if it's good. 

As for the second problem I made `def copy_tags` create Categories and Equipment from the original repository.

### DIFF
--- a/app/controllers/makes_controller.rb
+++ b/app/controllers/makes_controller.rb
@@ -5,8 +5,8 @@ class MakesController < SessionsController
 
   def create
     @repo = @repository.makes.build do |r|
-      r.title = params[:repository][:title]
-      r.description = params[:repository][:description]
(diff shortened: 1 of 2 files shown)


app/controllers/makes_controller.rb
```diff
@@ -44,10 +44,12 @@ def create_photos
     end
 
     def copy_tags
```
tahnok: This method isn't very accurate anymore... Maybe it should be renamed?

KarimChukfeh: done

PR #74, opened 2017-07-21
Rake task for brunsfield training spreadsheet

KarimChukfeh:
Parami shared with me a spreadsheet with 2100 records of training for people (mostly university students).

In this branch I created a rake namespace called csv which contains tasks that check the legibility of certifying these users based on. To be legible they have to:

1. Exist as a makerepo user
2. The type of training they got exists in makerepo


tahnok:
Seems like a good start, but I thought we had to make new users?

(also I think it's eligible, not legible)

lib/tasks/csv.rake
```diff
@@ -0,0 +1,42 @@
...
+      email = row[6]
+      training = row[10]
+      trainer = row[13]
+      course = row[14]
+
+      user = User.find_by(email: email)
+
+      if user && training && trainer
+        certs = user.certifications.map(&:training)
+        unless certs.include?(training)
```
tahnok: Is this going to work? Won't this return Training objects where as the training you have here is just the name?

lib/tasks/csv.rake
```diff
@@ -0,0 +1,42 @@
+namespace :csv do
+
+  task :certify_existing_users => :environment do
+
+    test_spreadsheet = Rails.root.join('test','lib','assets','TRAINING_DATABASE.csv')
```
tahnok: This should possibly also take an argument to a path on the disk

lib/tasks/csv.rake
KarimChukfeh: this works because the trianing i have isn't an objects. It's just the name obtained through the certification's training_session training name

PR #75, opened 2017-07-25
Add course option

KarimChukfeh:
This branch is to add the option to select and change the course field in staff's training session views


app/views/staff/training_sessions/show.html.erb
```diff
@@ -1,12 +1,16 @@
...
+<div>Course: <%= @current_training_session.course %></div><br>
 
 <div>Training Type: <%= @current_training_session.training.name %>
     <br>
     <%= form_for staff_training_session_path(@current_training_session), method: 'PATCH', action: 'update' do |f| %>
       <% label_tag 'changed_params[training_id]' %>
       <%= select_tag 'changed_params[training_id]', options_from_collection_for_select(Training.all, 'id', 'name'), prompt: 'change training' %><br>
+
+      <% label_tag 'changed_params[course]' %>
+      <%= select_tag 'changed_params[course]',  options_for_select(['GNG1105', 'GNG2205']), prompt: 'change course' %><br>
```
tahnok: I would suggest that the options for this be present in `TrainingSession` as a method.

tahnok:
This is good to merge

PR #77, opened 2017-07-26
Projects

Parastoo-S:
Projects


app/controllers/sessions_controller.rb
```diff
@@ -62,6 +64,10 @@ def update_activity_time
     session[:expires_at] = 72.hours.from_now
   end
 
+  def authorized_repo_ids
```
Parastoo-S: @tahnok Is this a good approach? I works!

tahnok: Ya, this looks exactly like what I had in mind!

Parastoo-S: Awesome:D

app/controllers/repositories_controller.rb
```diff
@@ -2,14 +2,29 @@ class RepositoriesController < SessionsController
...
+  before_action :check_auth
 
   def show
-    @photos = @repository.photos.first(5)
-    @files = @repository.repo_files.order("LOWER(file_file_name)")
-    @categories = @repository.categories
-    @equipments = @repository.equipments
-    @comments = @repository.comments.order(comment_filter).page params[:page]
-    @vote = @user.upvotes.where(comment_id: @comments.map(&:id)).pluck(:comment_id, :downvote)
+    if @repository.share_type.eql?("private")
```
tahnok: This should use your `private?` method in Repository

app/controllers/repositories_controller.rb
```diff
@@ -2,14 +2,29 @@ class RepositoriesController < SessionsController
...
+          redirect_to password_entry_repository_path
+      end
+    else
+      @photos = @repository.photos.first(5)
+      @files = @repository.repo_files.order("LOWER(file_file_name)")
+      @categories = @repository.categories
+      @equipments = @repository.equipments
+      @comments = @repository.comments.order(comment_filter).page params[:page]
+      @vote = @user.upvotes.where(comment_id: @comments.map(&:id)).pluck(:comment_id, :downvote)
+    end
```
tahnok: You can probably restructure a lot of this so that it has a single early return statement instead of 3 control paths as it is now. Something like

```
if @repository.private? && !check_passed
 redirect_to password_entry_repository_path and return #this last part is the bit that returns early
end

# existing code goes here
```

app/controllers/search_controller.rb
```diff
@@ -4,13 +4,13 @@ class SearchController < SessionsController
   require 'will_paginate/array'
 
   def explore
-    @repositories = Repository.paginate(:per_page=>12,:page=>params[:page]).order([sort_order].to_h).page params[:page]
+    @repositories = Repository.paginate(:per_page=>12,:page=>params[:page]).where(share_type: "public").order([sort_order].to_h).page params[:page]
```
tahnok: This `where(share_type: "public)` should be a scope on the model, although I think it should be the default scope, which I think you can do.

app/models/repository.rb
```diff
@@ -25,6 +28,12 @@ def self.license_options
     presence:   { message: "Project title is required."},
     uniqueness: { message: "Project title is already in use.", scope: :user_username}
 
+  validates :share_type,
```
tahnok: You should also validate that the share_type is included in a list of allowed types (eg: public or private)

app/controllers/repositories_controller.rb
Parastoo-S: oh ok

app/models/repository.rb
Parastoo-S: alright will do that

app/controllers/repositories_controller.rb
Parastoo-S: done

app/models/repository.rb
Parastoo-S: done

app/controllers/search_controller.rb
Parastoo-S: so, should I add a scope?

app/controllers/repositories_controller.rb
```diff
@@ -2,8 +2,13 @@ class RepositoriesController < SessionsController
   before_action :current_user
   before_action :signed_in, except: [:index, :show, :download, :download_files]
   before_action :set_repository, only: [:show, :add_like, :destroy, :edit, :update, :download_files]
+  before_action :check_session
+  before_action :check_auth
```
tahnok: `before_action` takes a list of actions that looks a bit cleaner. eg: `before_action :current_user, :check_auth, check_session`

app/controllers/repositories_controller.rb
```diff
@@ -104,14 +111,50 @@ def add_like # MAKE A LIKE CONTROLLER TO PUT THIS IN
...
   private
 
+    def check_session
+      if authorized_repo_ids.include? params[:id]
+        @authorized = true
+      end
+    end
+
+    def check_auth
+      if (@authorized == true || (@user.role == "admin") || (@user.role == "staff") || (params[:user_username] == @user.username))
```
tahnok: Isn't there a `.staff?` method on user?

app/controllers/repositories_controller.rb
```diff
@@ -182,4 +225,24 @@ def create_equipments
         Equipment.create(name: e, repository_id: @repository.id)
       end if params['equipments'].present?
     end
+
+    def update_share_type
```
tahnok: Do you need this method? I thought that ` @repository.update(repository_params)` would do this already?

app/controllers/search_controller.rb
tahnok: yes please

app/controllers/users_controller.rb
```diff
@@ -58,11 +58,19 @@ def change_password
...
+    if (params[:username] == @user.username)
+      @repo_user = User.find_by username: params[:username]
+      @github_username = Octokit::Client.new(access_token: @repo_user.access_token).login
+      @repositories = @repo_user.repositories.where(make_id: nil).page params[:page]
+      @makes = @repo_user.repositories.where.not(make_id: nil).page params[:page]
+      @photos = photo_hash
+    else
+      @repo_user = User.find_by username: params[:username]
+      @github_username = Octokit::Client.new(access_token: @repo_user.access_token).login
+      @repositories = @repo_user.repositories.where(share_type: "public").where(make_id: nil).page params[:page]
```
tahnok: If `@repositories` is the only variable difference here I would put just that in the if/else block and have everything else not repeated

app/models/repository.rb
```diff
@@ -1,5 +1,8 @@
 class Repository < ActiveRecord::Base
+  include BCrypt
+
   belongs_to :user
+  has_many   :users
```
tahnok: left over?

app/models/repository.rb
tahnok: I think there's a shortcut for this where you can say `validate :foo, inclusion: ['this', 'that']`

app/models/repository.rb
Parastoo-S: Oops yes

app/controllers/repositories_controller.rb
Parastoo-S: yes because to update the password, my conditions require the share_type to be saved. Also, for the "if @repository.update(repository_params)" to pass, we need the password to be in place. otherwise it will be false and will got to else statement. I figured this was the only way that everything could work. To save the share_type in order to update the password, then the if statement passes correctly. Hope this made sense and if there is a better way to approach this, please let me know.

app/controllers/repositories_controller.rb
Parastoo-S: done

app/models/repository.rb
Parastoo-S: ah you're right. Thanks!

app/controllers/repositories_controller.rb
Parastoo-S: yes there is! fixed

app/controllers/users_controller.rb
Parastoo-S: done

app/models/repository.rb
```diff
@@ -10,6 +12,9 @@ class Repository < ActiveRecord::Base
   belongs_to :parent,   class_name: "Repository", foreign_key: "make_id"
   paginates_per 12
 
+  # default_scope { where(:share_type => "public") }
+  scope :public_repos, -> { where(share_type:"public") }
```
Parastoo-S: @tahnok  can I use this regular scope? The default scope also affects the 'show' for the repositories. So even the user cannot see his/her own repositories.

tahnok: Maybe? I think the question comes down to which is the better default.

If you did `Repository.all` should you get private repos? On the other hand, if you do `@user.repositories` should you _not_ get private repos?

app/controllers/repositories_controller.rb
tahnok: I'm not sure I understand. If it's private the password needs to be there for any update?

app/controllers/repositories_controller.rb
```diff
@@ -224,15 +223,8 @@ def create_equipments
...
-    def update_share_type
-      if repository_params['share_type'].present?
-        @repository.share_type = repository_params['share_type']
-        @repository.save
-      end
-    end
-
     def update_password
-      if @repository.share_type.eql?("public")
+      if repository_params['share_type'].eql?("public")
```
Parastoo-S: @tahnok I modified the conditions so I could get rid of the update_share_type method.
Does this look better?
The update_password method is necessary in cases where the repo changes from private to public, I need the password to be set to nil manually.

app/models/repository.rb
Parastoo-S: If the user is looking at his/her own profile, they need to see the private ones too. But if it's someone else looking at their profile, they should _not_ see the private repos. the problem is that with the default_scope, @user.repositories only returns the public repos.

Parastoo-S: I think I'll use the scope :public_repos, -> { where(share_type:"public") } and call it anywhere it's appropriate.

app/assets/javascripts/repositories.js
```diff
@@ -0,0 +1,16 @@
...
+}
+
+function hideField(){
+	$("#change_pass").hide();
+	document.getElementById("password").style.display = 'none'
+}
+function toggleField() {
+		var x = document.getElementById("password")
+		if (x.style.display === 'none') {
+			document.getElementById("password").style.display = 'block'
```
tahnok: Can you do `x.style.display = 'block'` instead of getting the element again?

app/controllers/repositories_controller.rb
```diff
@@ -104,14 +108,50 @@ def add_like # MAKE A LIKE CONTROLLER TO PUT THIS IN
...
+
+  def pass_authenticate
+    @auth = Repository.authenticate(params[:slug], params[:password])
+    @repository = Repository.find_by_slug(params[:slug])
+    respond_to do |format|
+      if @auth
+        @authorized = true
+        authorized_repo_ids << params[:id]
+        flash[:notice] = "Success"
+        format.html{redirect_to repository_path}
```
tahnok: Does this need an id for the repository?

app/controllers/repositories_controller.rb
```diff
@@ -104,14 +108,50 @@ def add_like # MAKE A LIKE CONTROLLER TO PUT THIS IN
...
+        format.json { render json: { role: :guest }, status: :ok }
+      else
+        @authorized = false
+        flash[:notice] = "Incorrect password. Try again!"
+        format.html { redirect_to password_entry_repository_path}
+        format.json { render json: @repository.errors, status: :unprocessable_entity }
```
tahnok: Does `@repository` have errors here? I thought you just fetched it

app/controllers/repositories_controller.rb
```diff
@@ -104,14 +108,50 @@ def add_like # MAKE A LIKE CONTROLLER TO PUT THIS IN
...
   private
 
+    def check_session
+      if authorized_repo_ids.include? params[:id]
+        @authorized = true
+      end
+    end
+
+    def check_auth
+      if (@authorized == true || (@user.admin?) || (@user.staff?) || (params[:user_username] == @user.username))
```
tahnok: Can this line be bypassed by a hacker? I can control `params[:user_username]` so that it matches my username. I think you need to check that `@user.username == @repository.user_username`

app/assets/javascripts/repositories.js
Parastoo-S: done

app/controllers/repositories_controller.rb
Parastoo-S: fixed

app/controllers/repositories_controller.rb
Parastoo-S: done

app/controllers/repositories_controller.rb
Parastoo-S: fixed

app/controllers/repositories_controller.rb
```diff
@@ -1,9 +1,13 @@
 class RepositoriesController < SessionsController
-  before_action :current_user
+  before_action :current_user, :check_session
   before_action :signed_in, except: [:index, :show, :download, :download_files]
-  before_action :set_repository, only: [:show, :add_like, :destroy, :edit, :update, :download_files]
+  before_action :set_repository, only: [:show, :add_like, :destroy, :edit, :update, :download_files, :check_auth, :pass_authenticate]
+  before_action :check_auth, only: [:show]
```
Parastoo-S: @tahnok is it okay if I limited :check_auth to 'show'?
Also I had to move it after set_repository cus the repository needs to exist to check the conditions.

test/controllers/repositories_controller_test.rb
```diff
@@ -2,19 +2,20 @@
...
-
-    assert_response :success, "Failed at creating the repository"
-    assert Repository.find_by(title: "Repository44").present?
-  end
+  # test "users can create a repository" do
+  #   session[:user_id] = User.find_by(username: "mary").id
+  #   session[:expires_at] = "Sat, 03 Jun 2020 05:01:41 UTC +00:00"
+  #
+  #   post :create, user_username:"mary", slug:"repository44",
+  #   repository: {id: 44,
```
Parastoo-S: I have to associate the fixture photos(:four) to this repository.

app/controllers/repositories_controller.rb
tahnok: Ya, that makes sense

PR #78, opened 2017-07-26
Small admin views additions

KarimChukfeh:
In this branch I added view changes if the signed in user is an admin. 

1. If the signed in user is an admin; they'll be able to choose and later change the instructor of a trainer session before and during that training session.

2. If the signed in user is an admin and visits a user's profile they're able to change that user's role. I added some styling to that.


app/views/staff/training_sessions/new.html.erb
```diff
@@ -5,6 +5,10 @@
   <% label_tag 'training_id' %>
   <%= select_tag 'training_id', options_from_collection_for_select(Training.all, 'id', 'name'), prompt: 'select training' %><br>
 
+  <% if @user.admin? %>
+    <% label_tag 'user_id' %>
+    <%= select_tag 'user_id', options_from_collection_for_select(User.where('role LIKE \'staff\' or role LIKE \'admin\''), 'id', 'name'), prompt: 'select instructor' %><br>
```
tahnok: I would add a scope to the user for this to select admin and staff. Also why do you need `LIKE`? Doesn't `=` work?

eg: `User.where(role: ['staff', 'admin'])`?

app/views/staff/training_sessions/show.html.erb
```diff
@@ -127,4 +133,10 @@
     <%= submit_tag "Update", class: 'save-button' %>
   <% end %>
 
+  <% if @user.admin? %>
```
tahnok: In addition to showing and hiding the button to submit, you should add checks to the controller to make sure the current user is an `admin`.

app/views/staff/training_sessions/new.html.erb
KarimChukfeh: yeah this is much better ^_^

PR #79, opened 2017-07-26
Admin training sessions controls

KarimChukfeh:
A Brunsfield requirement is for admins to be able to change view and access training sessions and change the instructor. 

This branch adds an Admin::TrainingSessions controller and view.


app/views/admin/training_sessions/index.html.erb
```diff
@@ -0,0 +1,175 @@
...
+        </tbody>
+      </table>
+
+</section>
+
+<script>
+
+  var sort_direction = 1;
+
+  function findSession() {
```
tahnok: What do these do?

KarimChukfeh: this is for sorting the table using the down arrow in each column

KarimChukfeh: findSession() is for the search field in that view. 

When admin starts typing anything (date, trainer name, training type, etc) it filters the table to match only the rows containing the search

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -2,6 +2,7 @@ class Staff::TrainingSessionsController < StaffAreaController
 
   before_action :current_training_session, only: [:show, :edit, :update, :certify_trainees, :destroy]
   before_action :changed_params, only: [:update]
+  before_action :verify_ownership, only: [:show]
```
tahnok: Shouldn't this be on everything except show? Like shouldn't this apply for delete and update too?

app/views/admin/training_sessions/index.html.erb
tahnok: This should probably not live in the view directly, but instead in the js file associated with this section

app/views/staff/training_sessions/show.html.erb
```diff
@@ -140,10 +140,9 @@
     <%= submit_tag "Update", class: 'save-button' %>
   <% end %>
 
-  <% if @user.admin? %>
```
tahnok: Probably worth keeping this, or modifying it so that only admin and the current trainer can see it

test/controllers/admin/training_sessions_controller_test.rb
```diff
@@ -0,0 +1,23 @@
...
+  setup do
+    session[:user_id] = User.find_by(username: "adam").id
+    session[:expires_at] = "Sat, 03 Jun 2025 05:01:41 UTC +00:00"
+    @request.env['HTTP_REFERER'] = admin_training_sessions_url
+  end
+
+  test "admin can update training session user_id (trainer)" do
+    patch :update, id: 1, training_session: {user_id: 1337}
+    assert_redirected_to :back
+    assert_equal flash[:notice], "Updated Successfully"
```
tahnok: It's worth checking that the training session was in fact updated

KarimChukfeh: done

app/views/staff/training_sessions/show.html.erb
KarimChukfeh: done

app/controllers/staff/training_sessions_controller.rb
```diff
@@ -43,7 +44,10 @@ def update
     @current_training_session.update(changed_params)
 
     if params['dropped_users'].present?
-        @current_training_session.users -= User.where(username: params['dropped_users'])
+      @current_training_session.users -= User.where(username: params[ 'dropped_users'])
+      User.where(username: params[ 'dropped_users']).each do |user|
```
tahnok: I believe this would blow up if the user couldn't be found?

KarimChukfeh: is it enough to say `if User.where(username: params[ 'dropped_users']).each do |user|` ?

tahnok: Whoops, I wrote this wrong. It meant this would blow up if the user didn't have a certification. So you only need the `cert.present?` not the `user.present?`

KarimChukfeh: done

PR #81, opened 2017-08-02
Upgrading to rails 5

KarimChukfeh:
Handled many deprecation warnings mostly related to the style of passing params in controller tests.


Gemfile
```diff
@@ -1,16 +1,21 @@
 source 'https://rubygems.org'
 ruby '2.3.1'
 
-gem 'rails', '~> 4.2'
-gem 'thin', '~> 1.6.3'
+gem 'rails', '~> 5'
```
tahnok: Can you include at least one minor revision here? eg: 5.0 in this case

bin/update
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
```
tahnok: What is this file for? I can't tell from reading it

config/cable.yml
```diff
@@ -0,0 +1,9 @@
+development:
```
tahnok: I don't think we want this? I believe it's from ActionCable and it's wrong in production

config/environments/development.rb
```diff
@@ -27,50 +42,13 @@
...
-  config.assets.raise_runtime_errors = true
+  # Suppress logger output for asset requests.
+  config.assets.quiet = true
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
 
-  config.console = Pry
-
-  Octokit.configure do |c|
```
tahnok: Why were the following sections deleted?

config/puma.rb
```diff
@@ -0,0 +1,47 @@
+# Puma can serve each request in a thread from an internal thread pool.
+# The `threads` method setting takes two numbers a minimum and maximum.
+# Any libraries that use thread pools should be configured to match
+# the maximum value specified for Puma. Default is set to 5 threads for minimum
+# and maximum, this matches the default thread size of Active Record.
+#
+threads_count = ENV.fetch("RAILS_MAX_THREADS") { 5 }.to_i
```
tahnok: Where did this come from? I see that we have included `thin` which I believe is an alternative to `puma` for the webserver?

bin/update
KarimChukfeh: Many of these files were generated after running `rake rails:update`.  It was interactive and I was able to say `n` perhaps I should've done that?

[see 1.4 in the guide to upgrading Rails](http://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#the-update-task)

Gemfile
KarimChukfeh: done

config/cable.yml
KarimChukfeh: I don't fully understand what's wrong. Should I delete the file or just the first 3 lines?

config/environments/development.rb
KarimChukfeh: automatic file generation. adding it back

config/puma.rb
KarimChukfeh: deleted

config/cable.yml
KarimChukfeh: deleted

tahnok: Should probably still delete this if we're not using ActionCable

config/puma.rb
tahnok: Seems like it's still here?

KarimChukfeh: woops 😅

config/cable.yml
KarimChukfeh: I don't think it's there anymore

tahnok:
I think this branch can wait for now. There are more important things to do before september comes. Maybe a few weeks into the semester we can return to this?

KarimChukfeh:
check rails_5_branch instead

PR #82, opened 2017-08-02
Staff area

KarimChukfeh:
Staff area will be where staff navigate to different staff things (new training session, previous training sessions, kiosk)


config/routes.rb
```diff
@@ -118,6 +118,8 @@
     get 'index', path: '/'
 
     resources :training_sessions do
+      get 'index', path: '/', on: :collection
+      get 'new'
```
tahnok: Doesn't new already exist as a path?

PR #88, opened 2017-08-03
Reports Continued

Parastoo-S:
Fixing some old reports and adding new reports


app/controllers/admin/report_generator_controller.rb
```diff
@@ -44,4 +44,10 @@ def gender_frequency
     end
   end
 
+  def training
+    respond_to do |format|
+      format.html
+      format.csv {send_data ReportGenerator.training_report, filename: "gender_frequency-#{Date.today}.csv"}
```
tahnok: This looks like the wrong filename

app/models/repository.rb
```diff
@@ -33,6 +33,13 @@ def self.license_options
     self.user.decrement!(:reputation, 25)
   end
 
+  #GET VALUES FROM DATE PICKER (MAYBE)
+  # scope :between_dates_picked, ->(start_date , end_date){ where('created_at BETWEEN ? AND ? ', start_date , end_date) }
+
+  # Is this a good approach to generate the url?
```
tahnok: I think you can do ` Rails.application.routes.url_helpers.repository_path(slug: slug, user_username: user.username)`

app/controllers/admin/report_generator_controller.rb
Parastoo-S: fixed

Parastoo-S:
@tahnok So I managed to figure out how to show repositories according to selected dates. Do u want me to do this for every report? It might be beneficial if they want to look at the reports for the past semester for example. 
And I think it would be a quick fix. Just need to change _'.in_last_month'_ I to _'.where('created_at BETWEEN ? AND ? ', start_date, end_date)'_

tahnok:
Yes I think you should update the rest of the reports to allow selectable dates

app/controllers/admin/report_generator_controller.rb
```diff
@@ -1,47 +1,139 @@
...
+                      params[:end][:day].to_i)
+    @end_date = Time.zone.at(@end.to_time).to_datetime
+
+    selected_dates << @start_date << @end_date
+    redirect_to :back
+  end
+
+# New users
   def new_users
+    if session[:selected_dates] == nil
```
tahnok: One pattern that rubyist often rely on is that `nil == false`. That way can write `unless session[:selected_dates]` (because `unless` is the same as `if !`

app/controllers/admin/report_generator_controller.rb
```diff
@@ -3,7 +3,7 @@ class Admin::ReportGeneratorController < AdminAreaController
   require 'date'
 
   def index
-    if session[:selected_dates] == nil
+    unless session[:selected_dates]
```
Parastoo-S: @tahnok Like this?

PR #93, opened 2017-08-08
Fixing password-reset 

Parastoo-S:
The list of programs are not updated for when users want to update their info

### DIFF
--- a/db/migrate/20170809153751_update_old_users_required_fields.rb
+++ b/db/migrate/20170809153751_update_old_users_required_fields.rb
@@ -1,0 +1,7 @@
+class UpdateOldUsersRequiredFields < ActiveRecord::Migration
+  def change
+    User.where(name: nil).update_all(name: "anonymous")
+    User.where(gender: nil).update_all(gender: "unknown")
+    User.where(identity: nil).update_all(identity: "unknown")
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -11,9 +11,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-
 ActiveRecord::Schema.define(version: 20170823161336) do
-
 
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
@@ -146,8 +144,6 @@ ActiveRecord::Schema.define(version: 20170823161336) do
     t.integer  "make_id"
     t.integer  "make",          default: 0
     t.string   "slug"
-    t.string   "share_type"
-    t.string   "password"
   end
 
   add_index "repositories", ["user_id"], name: "index_repositories_on_user_id", using: :btree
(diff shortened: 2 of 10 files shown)


Parastoo-S:
I think the issue with the password-reset should be resolved now

Parastoo-S:
The select2 (dropdown for programs) is not working. Still trying to fix it.

db/migrate/20170809153751_update_old_users_required_fields.rb
```diff
@@ -0,0 +1,7 @@
+class UpdateOldUsersRequiredFields < ActiveRecord::Migration
+  def change
+    User.where(name: nil).update_all(name: "anonymous")
+    User.where(gender: nil).update_all(gender: "unknown")
```
tahnok: It might make sense to go even further with this and move all users that don't fit in one of our current bucks into 'unknown'. That should probably be a different PR though

db/schema.rb
```diff
@@ -11,7 +11,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170717220935) do
+
+
+ActiveRecord::Schema.define(version: 20170809153751) do
```
tahnok: This looks weird?

Parastoo-S:
@tahnok At the moment even invalid users can change their password because of the migrations that updates the required fields of name, gender, and identity. 
So those fields are filled with some values that make the user valid.
but if that's not enough I'll remove the check for validations for password_reset.

db/schema.rb
```diff
@@ -11,7 +11,10 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 20170808162101) do
+
+
+
+ActiveRecord::Schema.define(version: 20170809153751) do
```
tahnok: Looks like you need to re-dump the schema here. password/share_type

test/controllers/users_controller_test.rb
```diff
@@ -99,4 +99,41 @@ class UsersControllerTest < ActionController::TestCase
...
+    assert_redirected_to settings_profile_path
+  end
+
+  test "old valid user should be able to update profile with valid inputs" do
+    session[:user_id] = User.find_by(username: "john").id
+    session[:expires_at] = "Sat, 03 Jun 2020 05:01:41 UTC +00:00"
+
+    patch :update, username: "john", user: {name: "John", gender: "Male", identity: "grad", student_id: 9876543, faculty: "arts", program: "Honours BA in English", year_of_study: "3"}
+    assert_equal 'Profile updated successfully.', flash[:notice]
+    assert_equal "John", User.find_by(username: "john").name
```
tahnok: Why not use a variable for User here?

db/schema.rb
Parastoo-S: done

test/controllers/users_controller_test.rb
Parastoo-S: modified

PR #95, opened 2017-08-10
make view changes and add slug map in controller

KarimChukfeh:
One of the tasks is to replace course related projects with GNG2101 and GNG1103. This PR adds the `home` view change and a small controller change in order to allow for that change to happen.


KarimChukfeh:
@tahnok will merge as soon as [pull#96](https://github.com/uOttawa-Makerspace/MakerSpaceRepo/pull/96) is approved

PR #96, opened 2017-08-10
Migration to replace cpr with gng

KarimChukfeh:
This migration goes over every repo with a `Course-related Project` category and creates an appropriate `GNG1103` or `GNG2101` category so that it shows up in category search. If the course can't be determined from the title a `Other Projects` category is created instead.


KarimChukfeh:
@tahnok is this ready?

db/migrate/20170818004747_create_gng_categories_for_crp_repos.rb
```diff
@@ -0,0 +1,24 @@
...
+      if course == 'GNG2101'
+        repo_gng_cat = Category.new(repository_id: category.repository.id, name: course, category_option_id: gng2101.id)
+        repo_gng_cat.save
+        category.destroy
+      elsif course == 'GNG1103'
+        repo_gng_cat = Category.new(repository_id: category.repository.id, name: course, category_option_id: gng1103.id)
+        repo_gng_cat.save
+        category.destroy
+      else
+        repo_gng_cat = Category.new(repository_id: category.repository.id, name: 'Other Projects', category_option_id: CategoryOption.find_by(name: 'Other Projects').id)
```
tahnok: You can make this `CategoryOption` a variable since you use it a bunch of times

db/schema.rb
```diff
@@ -144,6 +144,7 @@
     t.integer  "make_id"
     t.integer  "make",          default: 0
     t.string   "slug"
+    t.boolean  "featured",      default: false
```
tahnok: I think this is extra?

KarimChukfeh: fixed

db/migrate/20170818004747_create_gng_categories_for_crp_repos.rb
KarimChukfeh: done

PR #99, opened 2017-08-14
Clean featured repos

KarimChukfeh:
This PR is to implement "featured repositories" in category search. Admins can go to a repository and mark it as featured. Featured repositories are marked and can be filtered in category search.


app/controllers/search_controller.rb
```diff
@@ -1,13 +1,24 @@
...
-  # before_action :signed_in
+  before_action :category, only: [:featured]
   require 'will_paginate/array'
 
   def explore
     @repositories = Repository.paginate(:per_page=>12,:page=>params[:page]).order([sort_order].to_h).page params[:page]
     @photos = photo_hash
   end
 
+  def equipment
```
tahnok: Where did this come from?

KarimChukfeh: in show repository you can click on any of the equipment tags and it renders a search view with no results so I added this to make it work

app/controllers/search_controller.rb
```diff
@@ -35,12 +45,21 @@ def category
         order_by sort_arr.first, sort_arr.last
       end
     end
-    if category && name
+
+    if category && name && params['featured']
+      old_category_repos = Repository.where(category: category, featured: true).distinct
+      new_category_repos = Category.where(name: name).distinct.includes(:repository).map(&:repository).select{ |repo| repo[:featured] == true }
+      @repositories = (old_category_repos + new_category_repos).uniq.sort!{|a,b|b.updated_at <=> a.updated_at}.paginate(:per_page=>12,:page=>params[:page]) do
```
tahnok: Don't need to `sort!` because 1) you aren't sorting in place, `sort` would work and 2) you can also do `sort_by(:updated_at)`

app/controllers/search_controller.rb
```diff
@@ -35,12 +45,21 @@ def category
...
+        order_by :updated_at, :desc
+      end
+
+    elsif category && name
```
tahnok: Would it make more sense to do something like

```
@repositories = (@repositories1 + @repositories2).uniq
if params[:featured]
 @repositories = @repositories.select{|r| r.featured?}
end

@repositories.paginate....
```

app/controllers/search_controller.rb
KarimChukfeh: cant use sort_by because `@repositories` is an array

app/controllers/search_controller.rb
KarimChukfeh: done

app/controllers/search_controller.rb
wesleyjellis: I think you mean you can't `order_by` on array's. [sort_by](https://ruby-doc.org/core-2.3.1/Enumerable.html#method-i-sort_by) is part of Enumerable

KarimChukfeh: it's weird I know but i keep getting 
> wrong number of arguments (given 1, expected 0)

on this

 >@repositories = @repositories.select{|r| r.featured?}.uniq.sort_by(:updated_at)

when sort_by is used

KarimChukfeh: ohh i was trying sort_by() all the time and it's actually sort_by{} with curly brackets 😅

PR #100, opened 2017-08-17
this should work according to capistrano3-unicorn github readme

KarimChukfeh:
[Under **Setup** in capistrano3-unicorn's readme](https://github.com/tablexi/capistrano3-unicorn) they say do 

> `namespace :deploy do`
> `  task :restart do`
> `    invoke 'unicorn:legacy_restart'`
> `  end`
> `end`

>if you need capistrano to cleanup your oldbin pid

not sure how to test this

### DIFF
--- a/config/deploy.rb
+++ b/config/deploy.rb
@@ -54,7 +54,7 @@ set :unicorn_config_path, 'config/unicorn.rb'
 after 'deploy:publishing', 'deploy:restart'
 namespace :deploy do
   task :restart do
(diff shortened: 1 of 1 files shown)


tahnok:
Hrmm or maybe it needs to be `    invoke 'unicorn:reload'`...

KarimChukfeh:
@tahnok I'm using `invoke unicorn:reload` now instead. 

Buut their documentation is confusing and I'm still not sure if we should use `legacy_restart` instead... :\

PR #101, opened 2017-08-17
compare lowercase attributes to lowercase query

KarimChukfeh:
In production when you go to Admin Area -> Manage Users and try to search for a user it doesn't always show all matching results. 

Parastoo suggests it's because in this line 

`@users = User.where('name like LOWER(?) OR email like LOWER(?) OR username like LOWER(?)', "%#
{@query}%", "%#{@query}%", "%#{@query}%")...`

the query compares the users' attributes, which usually contain uppercase letters like in `name`, with `LOWER(query)`. 

**Solution**
Compare `LOWER(name)`  (and other user attribute) to `LOWER(query)`


KarimChukfeh:
@tahnok this ready to merge?

PR #102, opened 2017-08-17
added controller action, front end, and tests

KarimChukfeh:
deals with the fact that users (including admins) can't delete their own (or any) comments

### DIFF
--- a/app/controllers/comments_controller.rb
+++ b/app/controllers/comments_controller.rb
@@ -7,7 +7,7 @@ class CommentsController < SessionsController
     comment = repository.comments.build(comment_params)
     comment.user_id = @user.id
   	comment.username = @user.username
-    	
+
   	if comment.save!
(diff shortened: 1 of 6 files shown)


app/controllers/comments_controller.rb
```diff
@@ -21,7 +21,23 @@ def create
 	  else
 	  	redirect_to root_path
 	  end
+  end
 
+  def destroy
```
tahnok: You can probably make this a bunch simpler by doing something like

```
if comment = Comment.find_by(id: params[:id])
  if @user.admin? || comment.user == @user
    @comment.destroy
     flash...
  else
   ...
  end
end
```

KarimChukfeh: done 😅

KarimChukfeh: done 😅

PR #105, opened 2017-08-18
Staff dashboard

KarimChukfeh:
Staff dashboard, or "kiosk", is where staff and admin can see who's in the space and easily add/remove users.

This PR includes a lot of view changes to make the user experience better.


app/models/training_session.rb
```diff
@@ -1,19 +1,32 @@
...
   validates :user, presence: { message: "A trainer is required"}
   validate :is_staff
+  validate :course_choice
 
   def is_staff
     errors.add(:string, "user must be staff") unless self.user.staff?
   end
 
+  def course_choice
+    if self.course == 'no course'
```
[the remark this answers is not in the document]
KarimChukfeh: done

app/views/staff_dashboard/index.html.erb
```diff
@@ -1 +1,146 @@
...
+<%= form_tag staff_dashboard_search_path, method: :get, action: :search do %>
+  <%= text_field_tag 'query', nil, placeholder: 'Search users by name, username, email...', class:'profile-text' %><br>
+  <%= submit_tag "Search", class: 'save-button' %>
+<% end %>
+</div>
+
+<h2><%= link_to "welcome to #{@space.name}", staff_dashboard_index_path%></h2>
+
+<div>
+  <%= form_tag("/staff_dashboard/change_space", method: "put", action: "change_space") do %>
```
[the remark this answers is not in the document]
KarimChukfeh: done

app/controllers/staff_dashboard_controller.rb
```diff
@@ -1,6 +1,76 @@
...
+      end
+    end
+    redirect_to :back
+  end
+
+  def sign_in_users
+    if params['added_users'].present?
+      params['added_users'].each do |username|
+        user = User.find_by(username: username)
+        lab_session = LabSession.new(user_id: user.id, sign_in_time: Time.now, sign_out_time: Date.tomorrow, mac_address: @space.pi_readers.first.pi_mac_address, pi_reader_id: @space.pi_readers.first.id)
```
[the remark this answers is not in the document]
KarimChukfeh: fixed

app/controllers/staff_controller.rb
```diff
@@ -1,4 +1,4 @@
-class StaffController < StaffAreaController
```
tahnok: Why this change? If there is some before action in `StaffDashboardController` that needs to be in more places, maybe it should be moved into `StaffAreaController`? Otherwise I fear that if StaffDashboardController will be the accidental implementation for some controller actions instead of causing an error

app/models/certification.rb
```diff
@@ -24,12 +30,19 @@ def self.to_csv (attributes)
   end
 
   def unique_cert
-    @user_certs = self.user.certifications
-    @user_certs.each do |cert|
-      if cert.training == self.training
-        errors.add(:string, "Certification already exists.")
+    @user_certs = self.user.certifications rescue nil
```
tahnok: I'm not a fan of these rescue statements because if catches _all_ exception, not just nil exceptions. See [this blog post](http://www.nickdesteffen.com/blog/why-rescue-nil-is-bad). You can do an `return true unless user`. Also, when who this happen that a certification has no user?

app/models/space.rb
```diff
@@ -5,10 +5,19 @@ class Space < ActiveRecord::Base
...
+
   validates :name,  presence: { message: "A name is required for the space"}, uniqueness: { message: "Space already exists"}
 
   def signed_in_users
-    return self.lab_sessions.where("sign_out_time > ?", Time.now).map(&:user)
+    return self.lab_sessions.where("sign_out_time > ?", Time.now).map(&:user).uniq
+  end
+
+  def recently_signed_out_users
+    users = self.lab_sessions.where(["(sign_out_time < ?) AND (sign_out_time > ?)", Time.now, 1.week.ago]).limit(40).map(&:user).uniq
```
tahnok: You probably want to order these by sign out time

app/controllers/staff_controller.rb
KarimChukfeh: `StaffDashbaordController` has the `before_action :default_space` which sets the space for the staff based on their latest tap. I can move that action to `StaffAreaController`

app/models/certification.rb
KarimChukfeh: This shouldn't happen but I didn't want anything breaking in the unlikely event that this happens

app/models/space.rb
KarimChukfeh: done

app/controllers/staff_area_controller.rb
```diff
@@ -10,4 +11,12 @@ def ensure_staff
     end
   end
 
+  def default_space
+    begin
+      @space = current_user.lab_sessions.last.space
+    rescue NoMethodError => e
```
tahnok: I'm really not a fan of rescue like this. You can use the [safe navigation operator](http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/) to do something like this `@space = current_user.lab_sessions&.last&.space || Space.first`

KarimChukfeh:
@tahnok these rescue clauses were there because in some model tests I test for presence of `user_id` and so on which return errors instead of failing without `rescue`. I've since removed the rescue clauses and updated the tests so that it doesn't do the checks that break it.

app/controllers/staff_area_controller.rb
KarimChukfeh: done

PR #107, opened 2017-08-18
Filtering invalid repositories

Parastoo-S:
In this PR I'm trying to solve the issue with the whole page breaking because of one invalid repository. 
I'm trying to add some custom validations to repository to indicate its validity and therefore filter those that are invalid and never show them in user's dashboard or search results.

This way, if a user has an invalid user in their dashboard for some reason. their whole profile page doesn't break.


test/fixtures/photos.yml
```diff
@@ -16,3 +16,10 @@ two:
   height: 3
   width: 3
 
+three:
+  repository_id: 3
+  image_file_name: p2
+  image_content_type: .png
+  image_file_size: 3
+  height: 3
+  width: 3
```
Parastoo-S: @tahnok What is a proper way of setting up fixtures for a photo? (also an array of photos)
I'm trying to see if these photos are valid. In the console I did ->
=> photo = Photo.find_by_repository_id(1)
then
=> photo.valid?
and returns false.
 I changed their file names to image.png, but that didn't solve the issue.

KarimChukfeh:
Does it make sense to have a default (valid) photo that we provide that shows up if a repository is invalid because it's images are invalid?

app/controllers/users_controller.rb
```diff
@@ -60,7 +60,7 @@ def change_password
   def show
     @repo_user = User.find_by username: params[:username]
     @github_username = Octokit::Client.new(access_token: @repo_user.access_token).login
-    @repositories = @repo_user.repositories.where(make_id: nil).page params[:page]
+    @repositories = @repo_user.repositories.has_photos.where(make_id: nil).page params[:page]
```
Parastoo-S: @tahnok Do u think this is a good approach to prevent invalid repos form breaking the user's profile for now?
(we might add default images later)
I can also add this to search.

Parastoo-S: nvm I don't like this cus it hides the repositories and it can cause confusion. I'll try the default image tmr

app/views/users/show.html.erb
```diff
@@ -142,8 +142,11 @@
   <% repos = params[:show].eql?('makes') ? @makes : @repositories %>
   <% repos.each do |repository| %>
     <div class="repository-container">
-
-      <div class="repo-display-wrapper" <%= "style=background-image:url('#{@photos[repository.id].image.url}')"  %> >
+      <% if repository.photos.exists? %>
```
Parastoo-S: I realized if there is no photo, the black background will show up for the repositories. So I didn't see a point in adding an icon or a default photo to them; instead I used the if statement to try to show the photo only if one exists.

Parastoo-S:
Is it ok if I added a NoImageAvailable image? I only used it for repository show.

Also for the user's profile, Karim suggested to use fa_icons to show up instead of the image but I decided to not replace the image with anything and leave the black background. is that alright?

If this approach is good (the if else statements) I will add them to search, explore and categories as well.

app/views/repositories/show.html.erb
```diff
@@ -18,6 +18,8 @@
         <!-- is this 'if p.valid?' necessary?(I added it to make sure an invalid photo doesn't cause error) -->
         <% if p.valid? %>
           <%= image_tag p.image.url, data: { width: p.width, height: p.height} %>
+        <% else %>
+          <%= image_tag("NoImageAvailable.png") %>
```
Parastoo-S: Or I can add another image that says "Photo Unavailable" or something. 
is it necessary to check the validity of photo here?

test/fixtures/photos.yml
tahnok: I think content_type might be something like 'image/png' or something (see https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type or https://en.wikipedia.org/wiki/Media_type

app/views/repositories/show.html.erb
```diff
@@ -9,19 +9,28 @@
...
       <div id="show-photo">
         <%= image_tag @photos.first.try(:image).try(:url), data: { width: @photos.first.width, height: @photos.first.height} %>
       </div>
-
-
     <div id="photo-slide">
       <% @photos.each do |p| %>
-        <%= image_tag p.image.url, data: { width: p.width, height: p.height} %>
+        <!-- is this 'if p.valid?' necessary?(I added it to make sure an invalid photo doesn't cause error) -->
+        <% if p.valid? %>
```
tahnok: I don't think so. The problem is usually repositories without images, not repositories with invalid images

app/models/repository.rb
```diff
@@ -25,6 +25,8 @@ def self.license_options
     presence:   { message: "Project title is required."},
     uniqueness: { message: "Project title is already in use.", scope: :user_username}
 
+  # validates :photos, :length => { :minimum => 1 }
```
tahnok: You could probably add this for new repos so that it's `validates :photos, :length => { :minimum => 1 }, on: :create` (or something like that)

test/fixtures/photos.yml
Parastoo-S: yess! thank youu

app/views/repositories/show.html.erb
Parastoo-S: oh ok

Parastoo-S: done

app/models/repository.rb
Parastoo-S: done

PR #110, opened 2017-08-23
Recieve rfid mac address

KarimChukfeh:
removed `try(:pi_location)` because `pi_reader.pi_location` is now being replaced with `pi_reader.space.name` using the `PiReader belongs_to :space` association.

### DIFF
--- a/app/controllers/rfid_controller.rb
+++ b/app/controllers/rfid_controller.rb
@@ -33,7 +33,7 @@ class RfidController < SessionsController
     if active_session.present?
       active_sessions.update_all(sign_out_time: Time.now)
       last_active_location = PiReader.find_by(pi_mac_address: active_sessions.last.pi_reader.pi_mac_address)
-      if active_location != new_location
+      if last_active_location != new_location
         new_session(rfid, new_location)
       end
     else
--- a/app/controllers/rfid_controller.rb
+++ b/app/controllers/rfid_controller.rb
@@ -31,8 +31,8 @@ class RfidController < SessionsController
     active_session = rfid.user.lab_sessions.where("sign_out_time > ?", Time.now)
     if active_session.present?
       active_session.update_all(sign_out_time: Time.now)
-      active_location = PiReader.find_by(pi_mac_address: active_session.first.try(:mac_address)).try(:pi_location)
-      new_location = PiReader.find_by(pi_mac_address: params[:mac_address]).try(:pi_location)
+      active_location = PiReader.find_by(pi_mac_address: active_session.first.try(:mac_address))
+      new_location = PiReader.find_by(pi_mac_address: params[:mac_address])
       if active_location != new_location
         new_session(rfid)
       end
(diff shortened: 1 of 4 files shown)


app/controllers/rfid_controller.rb
```diff
@@ -26,25 +26,27 @@ def card_number
...
   end
-  
+
   def check_session(rfid)
     active_session = rfid.user.lab_sessions.where("sign_out_time > ?", Time.now)
     if active_session.present?
       active_session.update_all(sign_out_time: Time.now)
-      active_location = PiReader.find_by(pi_mac_address: active_session.first.try(:mac_address)).try(:pi_location)
-      new_location = PiReader.find_by(pi_mac_address: params[:mac_address]).try(:pi_location)
+      active_location = PiReader.find_by(pi_mac_address: active_session.first.try(:mac_address))
```
tahnok: Can you do `active_session.pi_reader` instead?

app/controllers/rfid_controller.rb
```diff
@@ -26,25 +26,27 @@ def card_number
...
     end
   end
-  
+
   def new_session (rfid)
     sign_in = Time.now
     sign_out = sign_in + 3.hours
-    new_session = rfid.user.lab_sessions.new(sign_in_time: sign_in, sign_out_time: sign_out, mac_address: params[:mac_address])
-    new_session.save
+    if raspi = PiReader.find_by(pi_mac_address: params[:mac_address])
```
tahnok: Since you just found the PiReader in `check_session` you might as well pass it in as a parameter

app/controllers/rfid_controller.rb
KarimChukfeh: done

app/controllers/rfid_controller.rb
KarimChukfeh: done

app/controllers/rfid_controller.rb
```diff
@@ -26,25 +26,25 @@ def card_number
       end
     end
   end
-  
+
   def check_session(rfid)
     active_session = rfid.user.lab_sessions.where("sign_out_time > ?", Time.now)
+    new_location = PiReader.find_by(pi_mac_address: params[:mac_address])
```
tahnok: What should happen is no PiReader is found? I think you should return an error, but that breaks the old readers.... I think that's correct though

app/controllers/rfid_controller.rb
```diff
@@ -26,25 +26,25 @@ def card_number
       end
     end
   end
-  
+
   def check_session(rfid)
     active_session = rfid.user.lab_sessions.where("sign_out_time > ?", Time.now)
```
tahnok: I feel like this should be `active_sessions` since it's a list

app/controllers/rfid_controller.rb
KarimChukfeh: At the top of `def card_number` I create a new PiReader if no PiReader with `params[:mac_address]` is found. This makes adding new readers to spaces very easy

PR #111, opened 2017-08-23
Admin rfid and spaces manager

KarimChukfeh:
This PR adds changes to controller actions upon receiving information from Raspberry Pi rfid script. 

Now the controller expects the Pi's mac_address with each tap in order to properly sign a user into the corresponding space.

Finally, this PR adds changes for admins to be able to associate/disassociate a Raspberry Pi to a space and add/remove a training to a space

### DIFF
--- a/app/controllers/admin/spaces_controller.rb
+++ b/app/controllers/admin/spaces_controller.rb
@@ -3,7 +3,6 @@ class Admin::SpacesController < AdminAreaController
   layout 'admin_area'
 
   def index
-    @space = Space.first
   end
 
   def new
--- a/app/controllers/admin/spaces_controller.rb
+++ b/app/controllers/admin/spaces_controller.rb
@@ -27,10 +27,8 @@ class Admin::SpacesController < AdminAreaController
   end
 
   def destroy
-    if name = params[:name]
-      if Space.find_by(name: param).destroy
-        flash[:notice] = "Training removed successfully"
-      end
+    if Space.find_by(space_params).destroy
+      flash[:notice] = "Space deleted successfully!"
     else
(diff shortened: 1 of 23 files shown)


config/routes.rb
```diff
@@ -95,9 +95,18 @@
       end
     end
 
+    resources :spaces do
+      get 'index', path: '/', on: :collection
+      get 'edit', on: :member
```
tahnok: Do you actually need these? I though `resources :spaces` would create an index and an edit action. Also, if you aren't using some of the actions you should have `resources: :spaces, only: [...]` like here https://github.com/uOttawa-Makerspace/MakerSpaceRepo/blob/aec4840fd1dca34e9341c118fbbe6800d40498fc/config/routes.rb#L78

app/controllers/admin/spaces_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    space = Space.new(space_params)
+    if space.save
+      flash[:notice] = "Space created successfully!"
+    else
+      flash[:alert] = "Something went wrong."
+    end
+    redirect_to :back
+  end
+
+  def edit
```
tahnok: Isn't there an id param so you can not use :name and instead use :id ? I think it has to be since it will have been part of the path (eg: `/admin/spaces/123`)

config/routes.rb
KarimChukfeh: done

app/controllers/admin/spaces_controller.rb
KarimChukfeh: done

PR #113, opened 2017-08-26
User waiver

KarimChukfeh:


app/assets/javascripts/users.js
```diff
@@ -5,6 +5,16 @@ function hideField(){
 	document.getElementById("hidden").style.display = 'none';
 }
 
+function showCreateButton(){
+	if(document.getElementById("fake-signup-button").style.display != 'none'){
```
tahnok: You can use a variable here to avoid looking up the element every time

db/migrate/20170825180550_add_waiver_field_to_user.rb
```diff
@@ -0,0 +1,5 @@
+class AddWaiverFieldToUser < ActiveRecord::Migration
+  def change
+    add_column :users, :read_and_accepted_waiver_form, :boolean, default: false
```
tahnok: How is this different from Terms and Condition?

KarimChukfeh: After last meeting with Hanan we agreed that terms and conditions can be present at the footer of the website and not part of signing up. The waiver form will be replacing terms and conditions. 

The reason we need a new column because we have to make old users who agreed to Term and Conditions read the waiver form and agree to it.

app/assets/javascripts/users.js
KarimChukfeh: done

KarimChukfeh:
@tahnok thoughts on this PR?

app/views/users/new.html.erb
```diff
@@ -86,18 +86,223 @@
...
+<b>
+ON M’A RECOMMANDÉ ET DONNÉ L’OCCASION D’OBTENIR DES CONSEILS D’UN JURISTE IMPARTIAL
+AVANT DE SIGNER LE PRÉSENT DOCUMENT.</b>
+
+En cochant cette case, <b>J’ACCEPTE ET J’ASSUME PLEINEMENT</b> tous les risques et
+dangers pour la santé et la sécurité que peut comporter ma participation à
+l’Événement.
+</pre>
+                   <%= content_tag :span, @new_user.errors[:read_and_accepted_waiver_form].first, class: 'form-error' %>
+                   <%= f.radio_button :read_and_accepted_waiver_form, true, :onclick => "showCreateButton()", style:"display:inline;width:15px;"%>
```
Parastoo-S: doesn't having a checkbox make more sense than a radio button? cus if the person un-clicks this, the sign up button toggles but the radio button stays checked. 

Also, isn't it better to use enable, disable for the button instead of having two buttons? (This is just a thought I have to try it and see if it works. We'll talk about it)

app/controllers/users_controller.rb
```diff
@@ -114,7 +114,8 @@ def vote # MAKE A UPVOTE CONTROLLER TO PUT THIS IN
   def user_params
     params.require(:user).permit(:password, :password_confirmation, :url,
       :email, :name, :username, :avatar, :gender, :faculty, :use,
-      :description, :terms_and_conditions, :program, :student_id, :how_heard_about_us, :year_of_study, :identity)
+      :description, :terms_and_conditions, :program, :student_id, :how_heard_about_us,
```
Parastoo-S: Should we delete the terms_and_conditions column?

PR #118, opened 2017-08-31
Training reports

Parastoo-S:
> Admin and staff can download the information related to each training session and its users.

> Periodic emails of reports


app/controllers/staff/training_sessions_controller.rb
```diff
@@ -92,6 +92,14 @@ def destroy
     end
   end
 
+  def training_report
+    # binding.pry
```
tahnok: extra

PR #119, opened 2017-08-31
Let staff set rfid

KarimChukfeh:
This PR includes controller actions to `staff_dashboard_controller` to associate/dissociate an `Rfid` to a `User`.

Additionally, I removed links to edit_admin_user_url because:

a) the actions in the view are now available in other views under the Staff thing.
b) the update action is out of date and does things like delete all user certs (based on old certification concept)

so it's safest to remove access to that view even for admins

### DIFF
--- a/app/controllers/rfid_controller.rb
+++ b/app/controllers/rfid_controller.rb
@@ -14,8 +14,9 @@ class RfidController < SessionsController
         render json: { success: "RFID exist" }, status: :ok
         check_session(rfid)
       else
-        rfid.update(mac_address: params[:mac_address])
+        rfid.mac_address = params[:mac_address]
         rfid.touch
+        rfid.save
         render json: { error: "Temporary RFID already exists" }, status: :unprocessable_entity
       end
     else
--- a/app/controllers/rfid_controller.rb
+++ b/app/controllers/rfid_controller.rb
@@ -14,7 +14,8 @@ class RfidController < SessionsController
         render json: { success: "RFID exist" }, status: :ok
         check_session(rfid)
       else
-        rfid.update(params)
+        rfid.update(mac_address: params[:mac_address])
+        rfid.touch
         render json: { error: "Temporary RFID already exists" }, status: :unprocessable_entity
       end
     else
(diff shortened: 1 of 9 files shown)


app/controllers/rfid_controller.rb
```diff
@@ -14,6 +14,7 @@ def card_number
         render json: { success: "RFID exist" }, status: :ok
         check_session(rfid)
       else
+        rfid.update(mac_address: params[:mac_address])
```
tahnok: Why do `rfid.update` instead of just `rfid.mac_address = params[:mac_address]; rfid.save`? You can also drop the `touch` after this because updating the record will update the update_at which is what touch does

app/controllers/staff_dashboard_controller.rb
```diff
@@ -3,4 +3,23 @@ class StaffDashboardController < StaffAreaController
...
+      rfid.save
+    end
+    redirect_to :back
+  end
+
+  def unlink_rfid
+    if params['card_number'].present?
+      rfid = Rfid.find_by(card_number: params['card_number'])
+      rfid.user_id = nil
+      rfid.mac_address = Space.find_by(name: @user.location).pi_readers.first.pi_mac_address rescue rfid.mac_address
```
tahnok: What is this rescue for? are you trying to provide a default?

KarimChukfeh: i wanted the Rfid's mac address to change when the Rfid is dissociated from a user unless the Staff's location doesn't have any pi_readers

KarimChukfeh:
@tahnok yeah the current system was setup before the summer. Back then Staff didnt exist and only Admin was able set rfid to a user through Admin::UsersContoller

app/controllers/rfid_controller.rb
KarimChukfeh: i need `.touch` because if the mac_address is the same then `updated_at` won't update. 

>`test "posting existing card with no user updates updated_at" ` 
complains when i just do `.save` without `.touch` before it.

app/controllers/staff_dashboard_controller.rb
KarimChukfeh: fixed by using `if` instead

PR #121, opened 2017-09-01
rescue nil to avoid possible error on repo show

Parastoo-S:
There is a possibility of getting an error in controller show for repositories that don't have a photo.

### DIFF
--- a/app/controllers/repositories_controller.rb
+++ b/app/controllers/repositories_controller.rb
@@ -8,7 +8,7 @@ class RepositoriesController < SessionsController
     if @repository.private? && !@check_passed
       redirect_to password_entry_repository_path(@repository.user_username, @repository.slug) and return
     end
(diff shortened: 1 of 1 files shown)


app/controllers/repositories_controller.rb
```diff
@@ -8,7 +8,7 @@ def show
     if @repository.private? && !@check_passed
       redirect_to password_entry_repository_path(@repository.user_username, @repository.slug) and return
     end
-    @photos = @repository.photos.first(5)
+    @photos = @repository.photos.first(5) rescue nil
```
tahnok: I would prefer to not rescue nil here and instead do `@photos = @repository.photos&.first(5)`, possibly even `@photos = @repository.photos&.first(5) || []` so that you can still do `@photos.each` without hitting a nil error

Parastoo-S: done

PR #131, opened 2017-09-11
Training reports email schedule

Parastoo-S:
Hopefully the schedule will work properly now.


KarimChukfeh:
@Parastoo-S  I'm not sure if this will work or not. Maybe we can triple check with Wesley to make sure we're good?

Parastoo-S:
@tahnok Hey Wesley, can u please check if the syntax I wrote for the weekly email is correct?

PR #133, opened 2017-09-11
Remove rename Category

Parastoo-S:
Renaming a Category might create issues in sections of the website where category name is hard coded (ex. homepage categories)


Parastoo-S:
done

PR #139, opened 2017-09-13
no wrap of user names when hovered over

Parastoo-S:
when u hover over the training sessions' users, the names go to the next line. 
This small change should fix it.


KarimChukfeh:
thanks hahaha it was bothering me too!

PR #160, opened 2017-11-10
Gng categories

Parastoo-S:
Changing the Category names and adding the french section code.

### DIFF
--- a/app/controllers/search_controller.rb
+++ b/app/controllers/search_controller.rb
@@ -107,8 +107,8 @@ class SearchController < SessionsController
   SLUG_TO_CATEGORY_MODEL = {
    'internet-of-things' => 'Internet of Things',
    'course-related-projects' => 'Course-related Projects',
-   'gng2101' => 'GNG2101',
-   'gng1103' => 'GNG1103',
+   'gng2101/gng2501' => 'GNG2101/GNG2101',
+   'gng1103/gng1503' => 'GNG1103/GNG2503',
    'health-sciences' => 'Health Sciences',
    'wearable' => 'Wearable',
    'mobile-development' => 'Mobile Development',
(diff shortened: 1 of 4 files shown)


db/migrate/20171110152106_rename_gng_categories.rb
```diff
@@ -0,0 +1,6 @@
+class RenameGngCategories < ActiveRecord::Migration
+  def change
```
KarimChukfeh: maybe this should be `def up` so that it only runs once

app/controllers/search_controller.rb
```diff
@@ -107,8 +107,8 @@ def photo_hash
   SLUG_TO_CATEGORY_MODEL = {
    'internet-of-things' => 'Internet of Things',
    'course-related-projects' => 'Course-related Projects',
-   'gng2101' => 'GNG2101',
-   'gng1103' => 'GNG1103',
+   'gng2101/gng2501' => 'GNG2101/GNG2501',
+   'gng1103/gng1503' => 'GNG1103/GNG2503',
```
KarimChukfeh: Typo? GNG2503 should be GNG1503

db/migrate/20171110152106_rename_gng_categories.rb
```diff
@@ -0,0 +1,6 @@
...
+    Category.where(name: "gng1103").update_all(name: "gng1103/gng1503")
+    CategoryOption.where(name: "gng2101").update_all(name: "gng2101/gng1501")
```
KarimChukfeh: I'm confused as to what's happening here. Shouldn't this be `Category.where(name: "gng2101").update_all(name: "gng2101/gng2501")  `

Parastoo-S:
@KarimChukfeh I think I fixed all the mistakes. And made suggested changes.

PR #170, opened 2018-02-03
add sign out all users button

jimmyasyraf:
Justine suggested a sign-out-everyone button

### DIFF
--- a/test/controllers/staff_dashboard_controller_test.rb
+++ b/test/controllers/staff_dashboard_controller_test.rb
@@ -56,4 +56,10 @@ class StaffDashboardControllerTest < ActionController::TestCase
      assert_redirected_to :back
      refute users(:mary).rfid.present?
    end
+
+   test "staff can sign out everyone at once" do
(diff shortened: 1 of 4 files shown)


test/controllers/staff_dashboard_controller_test.rb
```diff
@@ -56,4 +56,10 @@ class StaffDashboardControllerTest < ActionController::TestCase
      assert_redirected_to :back
      refute users(:mary).rfid.present?
    end
+
+   test "staff can sign out everyone at once" do
+     session[:user_id] = users(:olivia).id
+     session[:expires_at] = "Sat, 03 Jun 2025 05:01:41 UTC +00:00"
+     get :sign_out_all_users
```
Parastoo-S: Good test case! I suggest add some assertion so you make sure your code does what u expect after you click the button (aka `get :sign_out_all_users`)
For ex. you can check if after the click, no signed_in_users are present in the space.
Also check if the pages gets redirected back to staff dashboard page

PR #179, opened 2018-07-20
180720 bm youtube link

brunomrlima:


app/views/repositories/new.html.erb
```diff
@@ -38,6 +38,10 @@
 
     <div id="image-container"></div>
 
+      <%= f.label :youtube_link, "Youtube Link" %>
+      <div>Your project will be better understood with a 1 min video Youtube.<br>Votre projet sera mieux compris avec une vidéo de 1 min Youtube</div>
```
Parastoo-S: 1 min Youtube video:)

app/views/repositories/show.html.erb
```diff
@@ -8,7 +8,11 @@
 
 <section id="show-repository">
 
+
   <section id="repository-photos">
+    <% if @repository.youtube_link %>
+      <%= youtube_video(@repository.youtube_link) %>
```
Parastoo-S: Is there a way to check if the provided link is in fact a youtube video? I'm thinking either a check that "youtube.com" is part of the provided link. but then if they provide a link for a channel it won't be valid as well. But I suggest taking a look into it.

Parastoo-S: Could you please also add the youtube_link field to the Edit for repositories? So they can edit it if they need to.

app/views/repositories/new.html.erb
brunomrlima: Thank you! Parastoo! I just corrected it ;)

app/views/repositories/show.html.erb
brunomrlima: Done!

brunomrlima: I'm using a gem that return nil if it's not a youtube video link.

Parastoo-S: Oh okay I see. that's good.

Parastoo-S: Great!

Parastoo-S:
Once thing I noticed is that if you go to edit and try deleting the YouTube URL, it actually does not delete the video from the repository. I think this is because you need to add `:youtube_url` to the `repository_params` in the Repository controller. This allows the update method to actually update the parameter of the targeted repository.

app/views/repositories/show.html.erb
```diff
@@ -8,7 +8,11 @@
 
 <section id="show-repository">
 
+
   <section id="repository-photos">
+    <% if @repository.youtube_link %>
```
Parastoo-S: I think writing `if @repository.youtube_link.present?` is what you would want to use here. When I tried deleting the url, it did not remove this section from the repository 'show' page. But try it for yourself after fixing the `update` and see which one works better.

brunomrlima: Well observed. Thanks

brunomrlima: Everything is working now.

PR #219, opened 2018-11-20
Updated selected gems

Parastoo-S:
GitHub has been alerting to update some of the gems found in the gem.lock file to protect our app from security vulnerabilities. 
I only updated the gems that was causing security vulnerability.

Security Alerts: https://github.com/uOttawa-Makerspace/MakerSpaceRepo/network/alerts

**Note**: Some other gems got updated as a by-product of updating the selected gems.


nicoco007:
Fixed in #247.

PR #1226, opened 2022-06-30
Removed Several Double Flash Notices

NomadNaomie:
Pages for repositories, editing profile and editing account settings all had an older form of flash notices/alerts, removed them so now the header controls them.


art29 (approved):
All good!

PR #1227, opened 2022-06-30
Adding support for multiple RFID boxes per space

NomadNaomie:
Change the sign in, sign out mechanism to lie on Space IDs instead of mac addresses. Added some unit testing as well, not sure if there is additional changes that need to be made to integrate it

### DIFF
--- a/spec/controllers/rfid_controller_spec.rb
+++ b/spec/controllers/rfid_controller_spec.rb
@@ -39,6 +39,54 @@ RSpec.describe RfidController, type: :controller do
         expected = { error: 'Temporary RFID already exists' }
         expect(actual).to eq(expected)
       end
+
+      it 'should not create rfid and should sign in RFID and sign out' do
+        pi_reader = create(:pi_reader)
+        attributes = create(:rfid, mac_address: pi_reader.pi_mac_address)
+        rfid_params = { rfid: attributes.card_number, mac_address: attributes.mac_address }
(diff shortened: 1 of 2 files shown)


spec/controllers/rfid_controller_spec.rb
```diff
@@ -39,6 +39,54 @@
...
+        sign_out_rfid_params = { rfid: attributes.card_number, mac_address: sign_out_pi.pi_mac_address }
+        expected = { success: 'RFID sign in' }
+        expect(actual).to eq(expected)
+        expect { post :card_number, params: sign_out_rfid_params, :format => :json }.to change(Rfid, :count).by(0)
+        actual = JSON.parse(response.body, symbolize_names: true)
+        expected = { success: 'RFID sign out' }
+        expect(actual).to eq(expected)
+      end
+
+      # #Temporary RFID Multi-box
```
art29, line 74: Is there a reason why there is a commented test?

art29 (changes requested):
Seems perfect to me! Only thing is the test that is commented.

spec/controllers/rfid_controller_spec.rb
NomadNaomie: I wasn't sure what the logic was for temporary RFID, whether they would need to be tested in a multi box setup or not

art29 (approved):

PR #1229, opened 2022-07-05
Fixing header switching issues. 

NomadNaomie:
Might be weird to move this ahead of all the other jQuery removal but the bug is on the site and is fixed in this code


art29 (approved):
Seems fine to me

PR #1234, opened 2022-07-17
Job Orders, Fixing Some JS

art29:


app/views/job_orders/new.html.erb
```diff
@@ -0,0 +1,36 @@
+<section x-data="{job_type_id: <%= @job_order.job_type_id.present? ? @job_order.job_type_id : JobType.first.id %>}">
+  <%= form_for(@job_order, url: {controller: "job_orders", action: "create"}) do |f| %>
+
+    <div class="text-center">
+
+      <h2>New Job Order (Step 1/4)</h2>
+
+      <div class="alert alert-info text-center">
```
NomadNaomie, line 8: Maybe an alert-warning or danger instead.

NomadNaomie:
This is a little visual nitpick but could we add an on change listener for the inputs that are money related and display them with fixed 2 decimal places, like $10.1 would show $10.10

app/javascript/controllers/comments-modal_controller.js
```diff
@@ -0,0 +1,24 @@
...
+    setCoHostContent(data) {
+        this.idTarget.innerHTML = data.id;
+        this.commentsTarget.innerHTML = data.comments.replace(/\n/g,'<br />');
+    }
+
+    open() {
+        document.body.classList.add("modal-open");
+        this.element.setAttribute("style", "display: block;");
+        this.element.classList.add("show");
+        document.body.innerHTML += '<div id="modal-backdrop" class="modal-backdrop fade show"></div>';
```
NomadNaomie, line 15: stylistic choice but to me its a bit easier to go back and make changes this was done via creating an element, setting its ID and classes and then appending modifying the HTML in code

app/controllers/job_orders_controller.rb
```diff
@@ -0,0 +1,396 @@
...
+          error = true
+        end
+      end
+    else
+      error = true
+    end
+
+    if error
+      redirect_to job_order_steps_path(@job_order, step: (params[:step].to_i - 1))
+      flash[:alert] = "An error occurred while saving the job order step. make sure that you uploaded the right file types. Please try again."
```
NomadNaomie, line 68: is filetype the only reason this block can fail? If we not we can convert error from bool to string and in each block set what the error would be and append it to the flash at the end if its not empty

app/assets/stylesheets/application.scss
```diff
@@ -188,4 +188,16 @@ div.form-error{
 
 .fc th {
   width: inherit;
+}
+
+.padding-responsive {
```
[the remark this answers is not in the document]
NomadNaomie: Additionally we might want to use a Bootstrap breakpoint definition instead of manually taking a value

app/views/job_orders/invoice.html.erb
```diff
@@ -0,0 +1,224 @@
...
+      .invoice-box table tr.item.last td {
+          border-bottom: none;
+      }
+
+      .invoice-box table tr.total td:nth-child(2) {
+          border-top: 2px solid #eee;
+          font-weight: bold;
+      }
+
+      @media only screen and (max-width: 600px) {
```
NomadNaomie, line 72: same as before, should we use BS breakpoints to keep things consistent?

art29: In this case, it's not really relevant since this will be converted to a PDF, I just followed guidelines I found online concerning emails and PDFs, I'll leave them that way just so they look the same as before

app/javascript/controllers/comments-modal_controller.js
art29: I'm not sure I understand what you are saying for this one

app/assets/stylesheets/application.scss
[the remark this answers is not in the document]
art29: I agree for the breakpoints, and this is something I use regularly for pages with not much contents like just a table and title for example

app/javascript/controllers/comments-modal_controller.js
NomadNaomie: something like this 

`let modalBackdrop = document.createElement('div')
modalBackdrop.classList.add("modal-backdrop");
modalBackdrop.classList.add("fade");
modalBackdrop.classList.add("show");
modalBackdrop.id = "modal-backdrop"
document.body.append(modalBackdrop)`

instead of 
`document.body.innerHTML += '<div id="modal-backdrop" class="modal-backdrop fade show"></div>'; `

I find it more readable and easier to make changes to

app/views/job_orders/new.html.erb
```diff
@@ -0,0 +1,36 @@
...
+        <i class="fa fa-warning"></i>
```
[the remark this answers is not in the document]
art29: Totally agree, what icon would you think fits best?

app/controllers/job_orders_controller.rb
art29: No it isn't, I'll try to improve this a bit

app/views/admin/shifts/shifts.html.erb
```diff
@@ -94,4 +94,10 @@
         padding-left: 3%;
         padding-right: 3%;
     }
-</style>
\ No newline at end of file
+</style>
+
+<script type="text/javascript" charset="utf-8">
```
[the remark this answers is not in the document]
art29: Yeah, I'll try to fix that soon, but I'll do it in another ticket

yarn.lock
```diff
@@ -2094,6 +2104,11 @@ browserslist@^4.0.0, browserslist@^4.12.0, browserslist@^4.14.5, browserslist@^4
     escalade "^3.1.1"
     node-releases "^1.1.71"
 
+bs-custom-file-input@^1.3.4:
```
NomadNaomie, line 2107: This is a BS4 package, are there BS5 alternatives or will we need to make something custom to handle what this handles?

NomadNaomie (commented):
Only thing now is the custom file input for me, everything else looks good

PR #1238, opened 2022-07-22
Fix: Broken Accordion Algorithm

art29:
The accordion algorithm to remember the state of the accordions wasn't working properly. It has now been fixed!


NomadNaomie:
Looks good to me, I am wondering if we wouldn't want to add a mechanism to override if an accordion is expanded by default. we could achieve this by instead of storing the list of IDs meant to be expanded, store an object with their state. If the ID isn't a key in the object then we allow its default either expanded or collapsed, if the key is in the object then we take the state. 

We'd have to Stringify and parse the Object every time to be able to store in localStorage but I don't think it adds that much overhead.

PR #1240, opened 2022-07-29
Feat: Job Orders Better Multiple File Upload

art29:
## Description
Better file upload for Job Orders when we can input multiple files

## Screenshot
image


app/views/job_orders/wizard/service.html.erb
```diff
@@ -10,7 +10,16 @@
         <div class="custom-file">
```
NomadNaomie, line 10: Should the line above be changed to something like 
`<p class="mt-0"><%= @job_order.user_files.attached? ? "Upload additional files here" : "Upload your file here" %></p>`
Just as an additional indicator that their previous files are already saved if they don't immediately scan for them below

PR #1243, opened 2022-08-08
Adding filtering for categories

NomadNaomie:
Adds a filter dropdown for search and explore pages, changed some controller code and added unit tests, let me know if there is a better approach


app/controllers/search_controller.rb
```diff
@@ -4,20 +4,37 @@ class SearchController < SessionsController
   before_action :current_user
 
   def explore
-    @repositories = Repository.paginate(per_page: 12, page: params[:page]).public_repos.order([sort_order].to_h).page params[:page]
+    if params[:category].blank?
+      @repositories = Repository.paginate(per_page: 12, page: params[:page]).public_repos.order([sort_order].to_h).page params[:page]
+    else
+      categories = Category.where(name: SLUG_TO_CATEGORY_MODEL[params[:category]]).where.not(repository_id: nil).distinct.includes(:repository).map(&:repository)
+      @repositories = Repository.paginate(per_page: 12, page: params[:page]).public_repos.where(id: categories.map(&:id)).order([sort_order].to_h).page params[:page]
```
art29, line 11: I think it could be simplified that way (To be double checked)
```rb
@repositories = Repository.paginate(per_page: 12, page: params[:page]).public_repos.includes(:categories).where({categories: {name: params[:category]}}).order([sort_order].to_h).page params[:page]
```

art29 (changes requested):
Seems mostly good! Fix the issue for French/English filters and try the improvement a gave for the Filtering (Would remove one line on each occurrence, it might not work tho)

art29 (approved):

PR #1245, opened 2022-08-13
Re-adding User Deletion

NomadNaomie:
Admins can now navigate to user pages and access the User Delete tab to delete users. They will be required to reauthenticate 

Delete Tab
image

Reauthentication Modal
image


app/controllers/admin/users_controller.rb
```diff
@@ -117,8 +117,12 @@ def delete_project_proposal
   end
 
   def delete_user
-    User.find(params[:id]).destroy
-    redirect_to root_path, notice: 'User Deleted!'
+    if User.authenticate(params[:username_email], params[:password]).admin?
```
art29, line 120: I would change this line to `if @user.admin? && @user.pword == password`. The reason being that this way the email doesn't need to be re-inputed.

app/views/users/show.html.erb
```diff
@@ -510,4 +524,27 @@
...
+        <button type="button" class="btn-close" data-bs-dismiss="modal"></button>
+      </div>
+      <div class="modal-body">
+          <div class="mb-3 attach">
+          <%= form_tag(delete_user_admin_users_path(id: @repo_user.id), method: :delete)  do %>
+            <span> Please re-enter your password to confirm your identity.</span>
+            <%= hidden_field_tag "id", @repo_user.id%>
+            <%= label_tag :delete_user, "Are you sure you want to delete this user?", class:"form-label mb-3" %>
+            <%= password_field_tag :password, nil, placeholder: t('sessions.login.password'), class: 'form-control mb-3' %>
+            <%= text_field_tag :username_email, params[:username_email], placeholder: t('sessions.login.username_or_email'), class: 'form-control mb-3', autofocus: true %>
```
art29, line 543: I would remove the email

art29 (approved):
Looks good, maybe just remove the email field for ease of use

NomadNaomie:
There are a lot of implications we've discussed, pending further discussions before we can move ahead with this.

art29 (approved):
Great Work on this! It works really well!

PR #1248, opened 2022-08-25
BS5 Button and Alert transparency fix

NomadNaomie:
Not a very clean fix, but until we can work out why .btn and .alert properties are being prioritized it works.


art29 (approved):
Don't like that fix very much but it is what is is. We'll find a better way later, maybe something about what gets imported first.

PR #1254, opened 2022-08-30
Feat: Added new annoucement type

art29:
Adding another type announcements that is visible to everyone and can be used for Maintenance Announcements for example


NomadNaomie (commented):
Replace All Users with All Visitors, replace Regular Users with All Users

NomadNaomie (approved):

PR #1258, opened 2022-09-01
Job order style multi file upload for repositories and project proposals

NomadNaomie:


art29 (changes requested):
Mostly good! Only issue is with the Drag & Drop, it goes under, which looks odd. If possible, try to add them over instead of under the Drag & Drop field.

NomadNaomie:
I can't add it under the drag and drop unfortunately, we can take out the drag and drop but this is what it looks like
image

app/controllers/repositories_controller.rb
```diff
@@ -98,7 +98,7 @@ def edit
 
   def create
     # @repository = @user.repositories.build(repository_params)
-
+    puts "repository_params: #{repository_params}"
```
art29, line 101: Probably remove that

app/javascript/packs/application.js
```diff
@@ -162,6 +162,21 @@ window.examResponse = function (exam_id, answer_id) {
     }),
   });
 };
+window.dragndrop = function (event) {
+  event.preventDefault();
+  let images = [...document.getElementsByClassName("image-upload")];
+  console.log(event.dataTransfer.items);
```
art29, line 168: Probably remove the console.log

art29 (approved):
You forgot a few debugging things, other than that, it's all good!

PR #1260, opened 2022-09-03
in-place role manager reloading

NomadNaomie:
Converted the staff and admin tables to partial, minor changes to the user controller response to trigger a js response (does not affect role manager tab on profile)


art29 (approved):
Works extremely well!

PR #1282, opened 2022-10-10
Task: Add badges to beginner from september 1st

art29:


lib/tasks/add_badges_for_beginner_trainings.rake
```diff
@@ -0,0 +1,78 @@
...
+    no_template_certifications = []
+    failed_certifications = []
+
+    Certification
+      .where(training_session_id: training_sessions_ids)
+      .each do |c|
+        badge_template = BadgeTemplate.find_by(training_id: c.training.id)
+
+        if badge_template.present?
+          updated_certifications << "#{c.user.name} - #{badge_template.badge_name} (#{b.acclaim_badge_id})"
```
NomadNaomie, line 22: B isn't defined, since the same thing is pushed into updated_certifications on 32 I assume we can get rid of this?

NomadNaomie (changes requested):

PR #1285, opened 2022-10-18
Feat: Cron Job to add/remove staffs from the Wiki

art29:


NomadNaomie (approved):
Since we won't be running it whenever a role is updated, looks fine to me

PR #1291, opened 2022-10-19
Using SSE to reload dashboards on RFID tap

NomadNaomie:
Ignore the branch name, it is using MessageBus

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -72,6 +72,7 @@ gem "syntax_tree", "~> 3.3.0"
 gem "syntax_tree-haml"
 gem "syntax_tree-rbs"
 gem "stimulus-rails"
+gem "message_bus"
 
 group :development, :test do
   gem "byebug", platform: :mri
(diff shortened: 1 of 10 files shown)


art29 (approved):
Looks great!
Only thing, would be there be a way to not re-render the tables if they haven't changed, not sure how we can do check that but might be cleaner that re-rendering them every 500ms? (If that is what it does, but from reading the code, it looks like it is doing that)

NomadNaomie:
> Looks great!
> Only thing, would be there be a way to not re-render the tables if they haven't changed, not sure how we can do check that but might be cleaner that re-rendering them every 500ms? (If that is what it does, but from reading the code, it looks like it is doing that)

It doesn’t, that’s the polling rate for checking for new messages from the server, no new messages no re-render and the messages are per space

art29:
> > Looks great!
> > Only thing, would be there be a way to not re-render the tables if they haven't changed, not sure how we can do check that but might be cleaner that re-rendering them every 500ms? (If that is what it does, but from reading the code, it looks like it is doing that)
> 
> It doesn’t, that’s the polling rate for checking for new messages from the server, no new messages no re-render and the messages are per space

All good then!

PR #1298, opened 2022-11-03
Making Requested Changes to MakerRoom

NomadNaomie:
- Calendar no longer scrolls
- Added Weekly or Monthly recurring meetings
- Fixed Automated Approval Switching
- Added a My Bookings Pane
- Merged Admin Panel into a Pane
- Changed display for bookings for admins, owner of the bookings and the public
- Added public bookings


app/controllers/admin/spaces_controller.rb
```diff
@@ -40,7 +40,20 @@ def create_sub_space
...
   def delete_sub_space
     if params[:id].present?
-      if SubSpace.find(params[:id]).delete
+      subspace = SubSpace.find(params[:id])
+      SubSpaceBooking
+        .where(sub_space: subspace)
+        .each do |booking|
+          status =
+            SubSpaceBookingStatus.find(booking.sub_space_booking_status_id)
+          status.sub_space_booking_id = nil
```
art29, line 49: Could we use `.update()` for these instead? It would save a line each time and is also easier to read

app/controllers/sub_space_booking_controller.rb
```diff
@@ -84,16 +149,65 @@ def bookings
...
+    unless signed_in?
+      redirect_to login_path, alert: "You must be logged in to view this page."
+      return
+    end
+    unless current_user.booking_approval || current_user.admin?
+      redirect_to root_path, alert: "You must be approved to book spaces."
+      return
+    end
+
+    if params[:sub_space_booking][:recurring].present?
```
art29, line 165: Idea: Maybe put that in a before_save or before_create in the model, more processing should be done on the model side, it would unclutter a bit the controller

app/controllers/sub_space_booking_controller.rb
```diff
@@ -219,11 +323,48 @@ def decline
...
     booking.booking_status_id = BookingStatus::DECLINED.id
     booking.save
-    redirect_to admin_sub_space_booking_index_path,
+    redirect_to sub_space_booking_index_path(anchor: "booking-admin-tab"),
                 notice:
                   "Booking for #{SubSpaceBooking.find(params[:sub_space_booking_id]).sub_space.name} declined successfully."
   end
 
+  def publish
+    unless current_user.admin?
```
art29, line 332: Create a method that checks if the user is an admin and add a before_action on the controller instead of doing it in multiple spots, it will be easier to read and save lines

app/javascript/packs/sub_space_booking.js
```diff
@@ -191,3 +222,104 @@ if (userSelect) {
     });
   }
 }
+document.addEventListener("turbolinks:render", ready);
+function ready() {
+  let anchor = window.location.hash.substring(1);
+  let pending_table = new URLSearchParams(window.location.search).get(
```
art29, line 228: If they aren't modified, `const` would be better

art29 (changes requested):
Added a few comments to improve the ease of reading the code, but the code itself is all good!

app/controllers/sub_space_booking_controller.rb
art29: Edit: I am wrong

app/controllers/sub_space_booking_controller.rb
```diff
@@ -84,16 +149,65 @@ def bookings
...
+            while start_time < end_date
+              params[:sub_space_booking][:start_time] = start_time + 7.days
+              params[:sub_space_booking][:end_time] = end_time + 7.days
+              start_time = params[:sub_space_booking][:start_time].to_datetime
+              end_time = params[:sub_space_booking][:end_time].to_datetime
+              book(params)
+            end
+          elsif params[:sub_space_booking][:recurring_frequency] == "monthly"
+            start_time = params[:sub_space_booking][:start_time].to_datetime
+            end_time = params[:sub_space_booking][:end_time].to_datetime
```
art29, line 183: Maybe merge the monthly and weekly together with a condition for the start and end times and save the result in a variable that could be used, it would save the duplication of lines

art29 (approved):

PR #1304, opened 2022-11-17
Volunteer Area Changes

NomadNaomie:
See Justine's message on teams.

Adding hide completed tasks, volunteer and staff multiple selections, clear input in tomselect, removal of creator from task


app/controllers/volunteer_tasks_controller.rb
```diff
@@ -101,16 +101,30 @@ def edit
...
+      User.where(
+        id:
+          VolunteerTaskJoin
+            .where(volunteer_task_id: @volunteer_task.id)
+            .where.not(user_type: "Volunteer")
+            .pluck(:user_id)
       )
+    @staff_available =
+      User
+        .where("users.role = ? OR users.role = ?", "staff", "admin")
```
art29, line 121: Would this work? `"users.role = staff OR users.role = admin"`, it would be better for SQL injection

art29 (approved):
Sorry for the delay, totally missed that PR, all good

PR #1307, opened 2022-11-21
Showing Language of Instruction, Level of training and course code for staff needed blocks

NomadNaomie:
MicrosoftTeams-image(1)
MicrosoftTeams-image


app/controllers/admin/shifts_controller.rb
```diff
@@ -273,8 +273,17 @@ def get_staff_needed
     SpaceStaffHour
       .where(space_id: @space_id)
       .each do |shift|
+        title = ""
+        shift.training_level_id == nil ?
+          title = "Staff Needed" :
```
art29, line 278: Maybe we could put that in the shift model? Might be slightly cleaner

art29 (approved):
Looks mostly good to me, just a small comment

app/models/space_staff_hour.rb
```diff
@@ -1,5 +1,7 @@
 class SpaceStaffHour < ApplicationRecord
   belongs_to :space
+  has_one :TrainingLevel
```
art29, line 3: Should be lowercase with underscores

app/models/space_staff_hour.rb
```diff
@@ -1,5 +1,7 @@
...
+  has_one :Course
```
art29, line 4: Should be lowercase with underscores

art29 (changes requested):

PR #1310, opened 2022-12-02
Feat/ics staff needed

art29:


NomadNaomie (approved):
We might want to take a look at the speed later when we're reviewing speeding up the rest of the calendar with the visibility toggling

art29:
> We might want to take a look at the speed later when we're reviewing speeding up the rest of the calendar with the visibility toggling

Yeah for sure, it clearly needs some optimization but at least it'll work

PR #1317, opened 2023-01-06
Front end design changes, moving await from Link_tos with methods

NomadNaomie:


art29 (approved):
Looks good, might want to update these two dropdowns tho: 
image

PR #1328, opened 2023-01-16
Detecting users with existing certifications in training sessions

NomadNaomie:


app/controllers/staff/training_sessions_controller.rb
```diff
@@ -156,7 +171,15 @@ def certify_trainees
     respond_to do |format|
       format.html do
         redirect_to staff_dashboard_index_path,
-                    notice: "Training Session Completed Successfully"
+                    notice:
+                      "Training Session Completed Successfully" +
+                        (
```
art29, line 176: would prefer `duplicates > 0 ? " (#{duplicates} duplicates skipped)" : ""`

art29 (approved):
Looks good to me, just a little nitpick

app/controllers/staff/training_sessions_controller.rb
NomadNaomie: So uh, prettier actually converts the ternary into that.

PR #1332, opened 2023-01-19
Adding dismiss to announcements

NomadNaomie:


art29 (approved):
Not really a fan of `Announcement.active.staff.where.not(id: AnnouncementDismiss.where(user_id: current_user.id).pluck(:announcement_id))` but we don't really have a choice since it is scoped to the user, so that's fine

PR #1340, opened 2023-01-31
Stopping duplicate shift creation when double clicking

NomadNaomie:


app/javascript/packs/admin_shifts.js
```diff
@@ -300,6 +300,7 @@ const populateUsers = (arg) => {
 
 // Calendar CRUD
 const createCalendarEvent = () => {
+  modalSave.disabled = true;
```
art29, line 303: Might be nice to add a loader as well? Bootstrap has some

art29 (approved):

PR #1341, opened 2023-01-31
Stop storing student numbers

NomadNaomie:
Student numbers are considered private, we do not use them so we will voluntarily delete them.


art29 (approved):
Please make sure **both** Justine and Alex approve this before merging it

PR #1342, opened 2023-02-01
Adding ability to copy shifts to next week

NomadNaomie:
There's a button to copy all shifts from the current week to the next week.

### DIFF
--- a/app/controllers/admin/shifts_controller.rb
+++ b/app/controllers/admin/shifts_controller.rb
@@ -123,6 +123,38 @@ class Admin::ShiftsController < AdminAreaController
     end
   end
 
+  def copy_to_next_week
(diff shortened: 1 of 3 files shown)


app/controllers/admin/shifts_controller.rb
```diff
@@ -123,6 +123,38 @@ def create
...
+        new_shift =
+          Shift.create(
+            space_id: @space_id,
+            start_datetime: new_start,
+            end_datetime: new_end,
+            reason: shift.reason,
+            pending: true
+          )
+        shift.users.each { |user| new_shift.users << user }
+        if !new_shift.save!
```
art29, line 145: I think it should be `!new_shift.save` otherwise it'll launch an error instead of redirecting (To be double checked)

art29 (approved):

PR #1343, opened 2023-02-01
Moving to Rails 7, Node 18, Vite

NomadNaomie:


app/javascript/entrypoints/forms.js
```diff
@@ -0,0 +1,813 @@
+// // function validation() {
```
art29, line 1: Wait, is that normal?

app/javascript/entrypoints/page_jumping.js
```diff
@@ -0,0 +1,28 @@
+// $(document).on("page:change", function () {
```
art29, line 1: Same here, is it normal?

app/javascript/entrypoints/photo_gallery.js
```diff
@@ -0,0 +1,45 @@
+// var photoArray = [];
```
art29, line 1: Same here, is it normal?

app/javascript/entrypoints/requests.js
```diff
@@ -0,0 +1,72 @@
+// $(document).on("page:change", function () {
```
art29, line 1: Do we use this?

art29 (changes requested):

PR #1344, opened 2023-02-04
adding password lockout

NomadNaomie:


app/controllers/sessions_controller.rb
```diff
@@ -41,15 +69,26 @@ def login_authentication
...
+            (
+              if user.auth_attempts > 1
+                "Incorrect password. You have #{User::MAX_AUTH_ATTEMPTS - user.auth_attempts} attempts left before your account is locked."
+              else
+                "Incorrect password."
+              end
+            )
+        else
+          error_message =
+            "Could not fin user with email or username #{params[:username_email]}"
```
art29, line 84: find

art29 (changes requested):
I think we want to add a way for admins to remove the lock before pushing that to prod, just to be safe

config/routes.rb
```diff
@@ -225,6 +225,8 @@
         delete "delete_project_proposal"
         delete "delete_user"
         patch "restore_user"
+        put "unlock_user"
+        put "lock_user"
```
art29, line 229: Since you use the unlock endpoint for both, you don't need the `lock_user`

art29 (approved):

PR #1345, opened 2023-02-06
Adding rate limiting, better password requirements, unlock via email

NomadNaomie:
- Rate Limiting with Rails Cache for login_authentication path
 - using the nobspw gem for better password protection with dictionary rejection
 - Adding an unlock email


art29 (approved):
Might want to keep the 8 characters, lowercase, uppercase and special character rule; To be double checked with Alex.

PR #1348, opened 2023-02-10
Email Typo Checking

NomadNaomie:
Using the levenshtein distance between the user supplied email against a domain list, we can detect typos in domains and supply corrections.
image


art29 (approved):
Nice work!

PR #1416, opened 2023-05-17
Fix/repo association mismatch

Columbium41:
## Description
Fixed AssociationTypeMismatch runtime error when creating a new repository with additional owners. Also made owner select fields to sort by username instead of name since username is unique while names can be duplicate.

### DIFF
--- a/app/controllers/repositories_controller.rb
+++ b/app/controllers/repositories_controller.rb
@@ -122,7 +122,9 @@ class RepositoriesController < SessionsController
 
       if params[:owner].present?
         params[:owner].each do |owner|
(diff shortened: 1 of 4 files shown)


app/controllers/repositories_controller.rb
```diff
@@ -122,7 +122,9 @@ def create
 
       if params[:owner].present?
         params[:owner].each do |owner|
-          @repository.users << User.find_by(username: owner)
+          if User.exists?(id: owner)
```
art29, line 125: A more rubylike syntax would be `if User.find(owner).present?`

art29 (approved):
Small change, rest seems good!

PR #1417, opened 2023-05-18
Fixed RFIDs not registering for users who haven't signed into spaces

Columbium41:
## Description
Fixed RFID unregister not working for users who have not signed into any spaces.


art29 (approved):
If you tested it and it works, all good!

PR #1419, opened 2023-05-23
Feat/sign up searchable programs field

Columbium41:
## Description
- Added a TomSelect that lets users search their programs in the sign up page and edit profile page
- Fixed a bug in sign up page where the `students-only` fieldset would not hide after being initially displayed
- Made placeholder text in sign up page more specific to improve accessibility

### Sign Up Page
sign-up-page

### Edit Profile Page
Screenshot 2023-05-23 at 2 52 06 PM


art29 (approved):
Nice work!

PR #1431, opened 2023-06-20
Fix/private repo password

Columbium41:
- Fixed a routing issue that would occur when entering an incorrect password in a private repository
- Fixed users being able to go to `/password_entry` in a public repo


art29 (changes requested):
Looks good but we need to fix the UI for the Repo Password Page, it's awful and old

art29 (approved):

PR #1436, opened 2023-06-26
Fix/broken repo comments

Columbium41:
- ## Description
- Fixed repository comments
     - Added scrollbar for really long comments and set a maximum height for comment containers
     - Added a 1000 character limit for comments to prevent users being able to enter extremely long comments
- Fixed comment upvoting/downvoting
     - Moved comment `vote` method from users controller to comments controller
     - Refactored `vote` method to redirect to the repository instead of a json render
     - Rewrote tests to match the new endpoint


app/controllers/comments_controller.rb
```diff
@@ -37,6 +44,71 @@ def destroy
                 )
   end
 
+  def vote
```
art29, line 47: If you can try to clean his method a bit, that would be nice, I know you just moved it, but it would be cleaner (for example, why do we have a rescue, maybe some stuff can be merged together to make it clearer etc.)

art29 (changes requested):
Looks good, tho a rework of the vote method would be very nice

art29 (approved):

PR #1441, opened 2023-07-05
Fix/profile joined projects

Columbium41:
## Description
- Fixed profile joined projects showing current logged in user's joined projects
- Refactored profile joined projects to look like repository cards instead of a table

## Screenshots
Screenshot_2023-07-05_at_3 12 10_PM


app/views/users/_repositories.html.erb
```diff
@@ -1,5 +1,9 @@
-<% if @repositories.empty? or params[:show].eql?('makes') and @makes.empty? %>
+<% if params[:show].eql?('repositories') && @repositories.empty? %>
```
art29, line 1: why not params[:show] == 'repositories' ?

app/views/users/_repositories.html.erb
```diff
@@ -45,23 +49,29 @@
...
-                    <td class="titlepp">
-                        <%= joined_project.project_proposal.title %>
-                    </td>
-                    <td>
-                        <%= link_to 'Show', joined_project.project_proposal, class: 'green-button' %>
-                    </td>
-                </tr>
-            </tbody>
-        </table>
+        <% project_proposal = joined_project.project_proposal %>
```
art29, line 52: Really not a fan of Ruby in HTML... For `has_user` you can create a method in the model

art29 (changes requested):
Few things to fix

app/views/users/_repositories.html.erb
Columbium41: I believe that was how it was done, so I continued to do it this way. I can change it to use double equals

art29 (approved):

PR #1442, opened 2023-07-06
Fix/admin certifications demotions

Columbium41:
## Description
- Fixed admin certification demotions page not displaying due to certain records without training sessions

## Screenshots
image


app/views/admin/certifications/_table_demotions.html.erb
```diff
@@ -15,15 +15,25 @@
 
     <tbody>
     <% @demotions.each do |demotion| %>
+      <% has_training_session = demotion.training_session.present? %>
```
art29, line 18: I would remove this and have the <% if demotion.training_session.present? %> go one level higher so we can put the <td> inside it and the else would just be 3 empty <td>

art29 (changes requested):
This needs a bit of rework

PR #1452, opened 2023-08-02
Task/rework profile navbar links

Columbium41:
## Description
- Removed some links in profile dropdown
     - Moved "Liked" to the explore repositories page
     - Removed "Settings"
     - Removed "Help"
     - Removed "Booking Admin" (confirmed with Justine)
- Removed "Buy CC" and "Volunteering" links from the navbar

#### Navbar Header
Screenshot 2023-08-02 at 12 21 43 PM

#### Repository explore page
Screenshot 2023-08-02 at 12 22 07 PM


config/locales/fr.yml
```diff
@@ -21,7 +21,7 @@ fr:
     create_project: "Documenter un projet"
     user:
       profile: "Profil"
-      liked: "J'aime"
+      liked: "Vos préférences"
```
art29, line 24: I think "Vos favoris" would be a better translation

art29 (changes requested):

PR #1453, opened 2023-08-03
Feat/recurring unavailabilities

Columbium41:
- Added `recurring (boolean)`, `start_datetime (datetime)`, `end_datetime (datetime)` columns to `StaffAvailabilities` model
- Added one-time unavailabilities in admin area and staff unavailabilities page 
- Fixed recurring unavailabilities displaying outside of the assigned time period
- Added an edit modal for editing unavailabilities in the admin area unavailabilities page
- Added modals for staff availabilities page


app/models/staff_availability.rb
```diff
@@ -2,12 +2,16 @@ class StaffAvailability < ApplicationRecord
...
   validates :time_period_id, presence: true
 
+  validates :start_time, :end_time, :day, presence: true, if: :recurring?
+  validates :start_datetime, :end_datetime, presence: true, unless: :recurring?
+
   def week_day
     Date::DAYNAMES[day.to_i]
   end
+
+  def recurring?
```
art29, line 14: Double check in rails docs but I believe this already exists for all the booleans automatically (might be wrong tho)

art29 (approved):
Small nit, but everything seems to work well! And the UI is pretty clean too! 

One thing I was thinking, we could talk about this with Alex but now that we have dates, maybe we should show all the time periods and default select the matching one in the dropdown in the modal? That way, we don't need to change the timeperiod if we want to access the next semester

art29 (changes requested):
Sorry, just checked one thing, everything else I said in the previous message applies but there's a timezone issue with the non-recurring events, they are taking the timezone into account and shifts the events

PR #1533, opened 2024-03-07
Fixed user demotion popup not showing up

solinm:


app/controllers/admin/spaces_controller.rb
Columbium41, line 1: I think this is for the max capacity feature

Columbium41 (approved):
LGTM

app/controllers/admin/spaces_controller.rb
solinm: Yeah there are some changes from max capacity within this commit, sorry

PR #1534, opened 2024-03-07
Staff availabilities redirects to admin availabilities if admin 

solinm:


app/controllers/admin/spaces_controller.rb
Columbium41, line 1: I think this is for another feature (the max capacity one)

Columbium41 (approved):
I think you included some code from the max capacity feature into this one but other than that it LGTM

app/controllers/admin/spaces_controller.rb
solinm: Should be fine to merge, not getting any merge conflict error with this PR and it's the same changes as the max capacity one

PR #1547, opened 2024-04-22
Made button on main page to join teams program

solinm:


app/views/static_pages/_programs.html.erb
Columbium41, line 1: I don't think the statuses match with the user.rb file. Since there are 8 cases to check for with only 3 conditions, it might be a lot easier to just check if the user is in each program separately and render based off that.

Columbium41 (changes requested):
Program statuses on static page show incorrect results. Example: If I join the dev & teams program it shows I am not in the dev program on the homepage

Columbium41 (approved):
LGTM

PR #1703, opened 2025-06-05
Feat/badge pages

valvinod:
The Badges Epic is about 50% complete.  The Badges, Badge_Templates, and Badge_Requirements tables have been dropped, and the necessary information has been moved to the Trainings and Certifications tables. Badge description pages have been added to the website to replace Credly's web pages, and all users with certifications have been awarded their corresponding badges on Makerepo. Currently, nonfunctional features include setting Training levels, starting proficient projects and checking for required badges when starting proficient projects. Unit tests need to be written for the functionalities added to Trainings and Certifications. Furthermore, the UI will be undergoing changes to accommodate for the detachment from Credly once the current changes have been reviewed and approved.


app/models/training.rb
```diff
@@ -10,10 +10,23 @@ class Training < ApplicationRecord
   has_many :learning_modules, dependent: :destroy
   has_many :shifts, dependent: :nullify
   belongs_to :skill, optional: true
+  has_many :badge_templates
 
-  validates :name, presence: true, uniqueness: true
+  validates :name_en, presence: true, uniqueness: true
```
PencilAmazing, line 15: Keep in mind, this column doesn't exist if the migrations never ran. And because there's migrations that run before this column is created, this line would cause an error because this column doesn't exist yet. A bit confusing, but that's why you keep DB actions atomic

app/views/admin/trainings/edit.html.erb
```diff
@@ -25,6 +27,33 @@
...
+          <%= f.check_box :has_badge %>
+          </div>
+
+          <div class="col-md-12 form-floating my-2">
+            <%= f.text_area :description_en, placeholder: 'Write an description...', class: 'form-control', required: true %>
+            <%= f.label 'Badge Description (English)', class: 'form-label' %>
+          </div>
+
+          <div class="col-md-12 form-floating my-2">
+            <%= f.text_area :description_fr, placeholder: 'Write an description...', class: 'form-control', required: true %>
```
PencilAmazing, line 43: write an description lol

app/views/badges/_badge_list.html.erb
```diff
@@ -3,38 +3,36 @@
...
 
-      <% if badges.badge_url.present? %>
-        <a style="width: 150px;" target="_blank" href="<%= badges.badge_url %>">
-          <%= image_tag(badges.badge_template.image_url, style: 'width: inherit;display:block; margin:0 auto;', alt: "Badge image: #{badges.badge_template.badge_name}") %>
-        </a>
-      <% else %>
-        <%= image_tag(badges.badge_template.image_url, :style => 'width: 150px; display:block; margin:0 auto;', 'data-bs-toggle' => 'tooltip', :title => "#{badges.issued_to} didn't accept the badge yet", :alt => "#{badges.issued_to} didn't accept the badge yet") %>
-      <% end %>
+          <a style="width: 150px;" class="mx-auto w-100!important d-block" target="_blank" href="./badges/<%= badges.id %>">
+            <%= render "badge_icon", width: 150, height: 150, margin: "auto", display: "block", level: badges.training_session.level, name_en: badges.training_session.training.name_en, name_fr: badges.training_session.training.name_fr, skill: badges.training_session.training.skill_id %>
```
PencilAmazing, line 25: The SVG partial should take a single `name` parameter and the value inputted should be localized instead

PencilAmazing: Note that you're passing `skill_id` not the skill level

PencilAmazing: and maybe I'm missing it, but the `margin` and `display` parameters aren't being used in the partial? They're passed as local variables, not html attributes

app/views/proficient_projects/show.html.erb
```diff
@@ -11,11 +11,11 @@
     </div>
   <% end %>
   <div class="text-center">
-    Badge that will be Awarded:
+    Training that will be Awarded:
```
PencilAmazing, line 14: I'd keep this as mentioning the badge

db/migrate/20250526194624_change_badge_name_from_badge_template.rb
PencilAmazing, line 1: Okay there's 21 migrations here, you could probably merge them all into one or two.

test/controllers/badges_controller_test.rb
PencilAmazing, line 1: We don't really use the test folder, we use the spec/ folder instead

app/views/badges/index.html.erb
```diff
@@ -14,15 +14,3 @@
 <div class="badge_list">
   <%= render 'badges/badge_list' %>
```
PencilAmazing, line 15: I somehow made this partial break. Not sure if it's your fault but here's a comment anyways :^)

PencilAmazing (changes requested):
Overall very good, I like the work

PencilAmazing (changes requested):
Minor fixes, but otherwise the branch seems good to me. I've done some quick tweaks here and there.

PR #1706, opened 2025-06-11
Fix/text overflow dev program

Maxson52:
ignore second commit which was to keep tab spacing consistent with rest of project. toggle "whitespace" in first commit to view actual styling changes.


Maxson52:
Good catch, should be good to go now. I only fixed the overflow issue for now... there is still a bit of wonkiness on smaller screens but nothing too crucial

PR #1711, opened 2025-06-20
Job orders revamp

Maxson52:
Refreshed the job orders system.
* UI refresh on client side, chat functionality between staff and clients. Clients have a more directed process for making a job order. All information is now centralized for an easier overview
* A new landing page.
* Redid admin side UI. Add job types and extra services to each job.
* Cleaned up the backend database schema


app/controllers/job_orders_controller.rb
```diff
@@ -545,8 +543,15 @@ def landing
   def update_job_order_with_comments(job_order, params)
     comments = params.delete(:comments)
     comment_from_staff = params.delete(:user_comments)
+    uploaded_files = params.delete(:user_files)
 
     success = job_order.update(params)
+ 
+    if success && uploaded_files.present?
```
Maxson52, line 550: can't figure out how to keep the previously uploaded files... somewhere they are getting deleted on form resubmission instead of persisting and/or being attached with newly uploaded files.

PR #1731, opened 2025-08-26
fix email quote not showing up, filter by completed not working

Maxson52:
as per justine's issues #513 and #514


Maxson52:
should be working but will result in a lot of emails... every chat instantly sends an email to alex, justine, an assigned staff, and the client (minus whoever send the message of course).

PR #1745, opened 2025-09-09
fix the sorting issue

UnknownCrafts:
recently signed out table is sorted to alphabetical names by default, now it should sort by last seen time by default.


PencilAmazing:
I believe this would affect all datatables and disable default sorting. I'd keep it, but then any sorting would have to happen server-side or via a HTML data attribute on the table

UnknownCrafts:
> I believe this would affect all datatables and disable default sorting. I'd keep it, but then any sorting would have to happen server-side or via a HTML data attribute on the table

Wouldn't it be better to sort in the frontend? I am not sure if that would also mean less database calls, which is the only reason I think that would be good.

PR #1746, opened 2025-09-10
Staging

PencilAmazing:


Maxson52:
> currently there's an issue with open hours not displaying on staging... will fix asap before merge to master

fixed in #1755

PR #1917, opened 2026-03-08
Feat/Added a page explaining CC moneys

UnknownCrafts:


PencilAmazing:
I'd remove the accordions, make them always open. Also needs to be french, I'll do that soon

